PR #6, opened 2016-06-29
Added bounding box as a scene environment option

favreau:


brayns/Brayns.cpp
```diff
@@ -192,6 +191,18 @@ struct Brayns::Impl
             _intializeExtensionPluginFactory( );
         _extensionPluginFactory->execute( );
 #endif
+        if( _parametersManager->getRenderingParameters().getSunOnCamera() )
```
bilgili: getSunOnCamera() ?

bilgili:
+1

PR #8, opened 2016-06-30
Fixed morphology loading - Diameter vs radius

favreau:


hernando:
Tests need fixing.

brayns/io/MorphologyLoader.cpp
```diff
@@ -201,7 +201,7 @@ bool MorphologyLoader::_importMorphology(
                 const float previousRadius =
                     (_geometryParameters.getRadius() < 0.f ?
                     -_geometryParameters.getRadius() :
-                    samples[ i - step ].w() * _geometryParameters.getRadius( ));
+                    samples[ i - step ].w() * 0.5f * _geometryParameters.getRadius( ));
```
hernando: Why not getRadiusFactor or getRadiusMultiplier?
Even better, you could add the function float correctRadius(float radius) in GeometryParameters to do this entire operation. This way you woudn't have to use a negative value to force an absolute value (which is a trick I'd rather avoid).

favreau: Good point! Thanks. Working on it.

PR #16, opened 2016-07-21
Introduced engines abstraction layer

favreau:
In order to be able to choose between OSPRay, Optix and Firerays as acceleration libraries, an abstraction of the ray-tracing engine is needed. This pull requests implements those new engine classes, as well as the factory that is to be used to instanciate them. The engine can be chosen using the --engine command line option.


brayns/Brayns.cpp
```diff
@@ -46,72 +44,44 @@ namespace brayns
...
+        if( !_engine )
+            throw std::runtime_error( "Unsupported engine: " + engineName );
 
         // set HDRI skybox if applicable
         const std::string& hdri =
             _parametersManager->getRenderingParameters().getHDRI();
+        ScenePtr scene = _engine->getScene();
         if( !hdri.empty() )
-            _scene->getMaterial(MATERIAL_SKYBOX)->setTexture(TT_DIFFUSE, hdri);
+            scene->getMaterial(MATERIAL_SKYBOX)->setTexture(TT_DIFFUSE, hdri);
```
bilgili: this is very engine specific ?

bilgili: Different engines may have different initializations.

brayns/Brayns.cpp
```diff
@@ -46,72 +44,44 @@ namespace brayns
...
+        scene->addLight( sunLight );
 
+        // Build geometry
         loadData( );
-        _scene->buildEnvironment( );
-        _scene->buildGeometry( );
-        _scene->commit( );
+        scene->buildEnvironment( );
+        scene->buildGeometry( );
+        scene->commit( );
```
bilgili: commit is also OSPray centric.

brayns/Brayns.cpp
```diff
@@ -137,16 +107,15 @@ struct Brayns::Impl
...
     }
 
     void render( const RenderInput& renderInput,
                  RenderOutput& renderOutput )
     {
         reshape( renderInput.windowSize );
 
-        _frameBuffer->map( );
-        _camera->set(
+        _engine->getFrameBuffer()->map( );
```
bilgili: map is OSPray centric. I think the initialization should be in the engine construction.

brayns/Brayns.cpp
favreau: No, this is not engine specific. The texture types (diffuse, normal, etc) are generic, as well as the ID for the material ID for the Skybox. It's the job of the engine to set the texture accordingly for the renderer.

brayns/Brayns.cpp
favreau: Well, not really. GPU engine such as Optix also need to transfer data from CPU to GPU memory and this is the job of the commit method. That will eventually also be useful for out-of-core rendering.

brayns/Brayns.cpp
favreau: I agree. Changing it. Thanks for that! :)

brayns/Brayns.cpp
```diff
@@ -267,47 +242,46 @@ struct Brayns::Impl
...
         const Vector3f diag   = worldBounds.getSize( );
         Vector3f position = target;
         position.z( ) -= diag.z( );
 
         const Vector3f up  = Vector3f(0.f,1.f,0.f);
-        _camera->setInitialState(position, target, up);
-        _camera->setAspectRatio(
-            static_cast< float >( _frameSize.x()) /
-            static_cast< float >( _frameSize.y()));
+        camera->setInitialState(position, target, up);
```
bilgili: looks like copy pasted code. normalize it !

plugins/engines/Engine.cpp
```diff
@@ -0,0 +1,32 @@
...
+ * along with this library; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include "Engine.h"
+
+namespace brayns
+{
+
+void Engine::setActiveRenderer( const std::string renderer )
```
bilgili: pass by ref string

plugins/extensions/plugins/DeflectPlugin.cpp
```diff
@@ -66,9 +67,11 @@ void DeflectPlugin::_initializeDeflect()
 
 void DeflectPlugin::_sendDeflectFrame()
 {
-    const Vector2i frameSize = _extensionParameters.frameBuffer->getSize();
+    FrameBufferPtr frameBuffer =
+        _extensionParameters.engine->getFrameBuffer();
+    const Vector2i frameSize = frameBuffer->getSize();
```
bilgili: const ref

plugins/engines/Engine.h
```diff
@@ -0,0 +1,93 @@
...
+       Commits changes to the engine. This include scene and camera
+       modifications
+    */
+    virtual void commit() = 0;
+
+    /** Renders the current scene and populates the frame buffer accordingly */
+    virtual void render() = 0;
+
+    /** Executes engine specific pre-render operations */
+    virtual void preRender() = 0;
```
bilgili: I think post and pre functions does not need to be pure. You can add an empty implementation. Not every engine may need them.

brayns/Brayns.cpp
favreau: Not sure if I understand what you mean there :-/ Can you explain a bit more?

plugins/engines/Engine.cpp
favreau: Indeed! Shame on me.

plugins/engines/Engine.h
favreau: Hum... The idea of forcing the implementation in the inherited classes is one way to remind the developper that those methods exist, and prevent him/her to put the code in the render() method instead. That's why I'd rather keep them pure, but I'd like to have your opinion on that :-)

plugins/extensions/plugins/DeflectPlugin.cpp
favreau: Yep... shame on me again :'(

brayns/Brayns.cpp
bilgili: There are multiple places the same code, can you gather into a function. Normalization means decreasing the repetition.

plugins/engines/Engine.h
bilgili: Developer should be aware of that. They are generally going to have a look at headers for implementation or example implementations.

bilgili:
+1

plugins/engines/Engine.h
favreau: Fair enough

PR #18, opened 2016-07-26
ZeroEQ/Buf Cleanup

favreau:


bilgili:
Dont forget to update change log. We can have more brief information about what you are doing.

tribal-tec:
where is the cleanup? Just a renaming, no?

favreau:
A bit of both renaming and cleaning, basically refactoring to get rid of a circular dependency, and CMake cleanup since current implementation simply does not work on MacOSX

PR #19, opened 2016-07-27
Updated changelog with latest updates

favreau:


doc/ChangeLog.md
```diff
@@ -3,9 +3,38 @@ Changelog {#Changelog}
 
 # master {#master}
 
+* [#19](https://github.com/BlueBrain/Brayns/pull/19):
+  Updated changelog with latest updates.
```
bilgili: You dont have to write every pull req :) Only important ones adding, changing behaviour.

bilgili:
+1 after the first change log item removed

PR #21, opened 2016-08-02
Refactored, improved and documentated ISPC renderers

favreau:


favreau:
retest this please

apps/BraynsViewer/BraynsViewer.cpp
```diff
@@ -69,7 +69,7 @@ void BraynsViewer::keypress(char key, const Vector2f& where)
             _timestampIncrement << std::endl;
         break;
     case 'x':
-        sceneParams.setTimestamp( 107.f );
```
bilgili: Magic values !

brayns/common/scene/Scene.cpp
```diff
@@ -244,19 +240,20 @@ void Scene::buildDefault( )
 
     // Lamp
     ++material;
+    const float y = 0.99f;
```
bilgili: why a variable ?

apps/BraynsViewer/BraynsViewer.cpp
favreau: Only for debugging purpose. But ok.

brayns/common/scene/Scene.cpp
favreau: Because the value is used several times in the following lines. To make it coherent, I defined the x and z values in the same way.

favreau:
retest this please

bilgili:
+1

PR #25, opened 2016-08-10
Use memory map file to store simulation data

favreau:


brayns/common/simulation/SimulationDescriptor.cpp
```diff
@@ -0,0 +1,119 @@
...
+    struct stat sb;
+    if( ::fstat( _cacheFileDescriptor, &sb ) == -1 )
+    {
+        BRAYNS_ERROR << "Failed to attach " << cacheFile << std::endl;
+        return false;
+    }
+
+    _memoryMapPtr = ::mmap(
+        0, sb.st_size, PROT_READ, MAP_PRIVATE, _cacheFileDescriptor, 0 );
+    if( _memoryMapPtr == MAP_FAILED )
```
bilgili: on destruct you may be unmapping an invalid _memoryMapPtr, check that out

hernando:
As discussed I understand you're going to remove the mmap, but just for the record, memory mapped files are implemented in Lunchbox.
It's contradicting that you want to keep dependencies low so people can easily compile Brayns everywhere, yet this code is going exactly in the opposite direction as it's not portable. As always, there are trade-offs and to me it's clear that in the C++ world portability needs third-party multiplatform libraries (unless you want to write Boost yourself).

favreau:
@hernando Now that I know more about the Brion implementation, it's definitely a possibility but I still need to compare how it performs, so I will leave it to another PR.
Regarding the portability, the current implementation was tested on Linux and Mac, which are the supported platforms for Brayns, so it's not really an issue for now. But thanks for the lunchbox tip, I did not know about it and will definitely have a look.

brayns/common/simulation/SimulationDescriptor.cpp
favreau: Not quite sure how :-/ Any case in particular?

favreau: Ok, got it! Invalided the _memoryMapPtr pointer in case of MAP_FAILED.

bilgili:
+1

PR #29, opened 2016-09-05
Add support for NEST circuits (and spikes visualization)

jafyvilla:


brayns/common/CMakeLists.txt
```diff
@@ -47,7 +45,7 @@ set(BRAYNSCOMMON_PUBLIC_HEADERS
 )
 
 set(BRAYNSCOMMON_LINK_LIBRARIES
-    PUBLIC braynsParameters Servus boost_filesystem vmmlib boost_system
+    PUBLIC braynsParameters Servus boost_filesystem vmmlib boost_system Brion
```
favreau: Only the IO library should depend on Brion. To be removed.

brayns/io/MorphologyLoader.cpp
```diff
@@ -25,7 +25,7 @@
 #include <brayns/common/geometry/Cylinder.h>
 #include <brayns/common/geometry/Cone.h>
 #include <brayns/common/scene/Scene.h>
-#include <brayns/common/simulation/SimulationDescriptor.h>
+#include <brayns/io/simulation/SimulationDescriptor.h>
```
favreau: Since they are used but the Scene class, simulation related classes must stay in the common lib, not in the IO.

favreau:
Merged into #31

PR #35, opened 2016-10-13
Fixed CMake issues and corresponding side effects

favreau:


.gitignore
```diff
@@ -6,10 +7,20 @@ Servus/
 ZeroBuf/
 build/
 ZeroEQ/
+Brion/
+Lexis/
+Lunchbox/
+MVDTool/
+httpxx/
+vmmlib/
```
hernando: You wouldn't need this if you were using -DCOMMON_SOURCE_DIR :)

.gitsubprojects
```diff
@@ -1,8 +1,14 @@
...
+
+# ZeroEQ and HTTP messaging
 git_subproject(ZeroBuf https://github.com/HBPVIS/ZeroBuf.git 346c1b5)
 git_subproject(ZeroEQ https://github.com/HBPVis/ZeroEQ.git 6de9a8a)
 git_subproject(Lexis https://github.com/HBPVis/Lexis.git 06819cc)
+
+# Streaming to display walls
 git_subproject(Deflect https://github.com/BlueBrain/Deflect.git 1cb94bc)
+
+# Data access
```
eile: This will be stripped by ```ninja Brayns-update-git-subprojects```

brayns/io/CMakeLists.txt
```diff
@@ -34,7 +33,7 @@ if(ASSIMP_FOUND)
 endif()
 
 if(TARGET Brion)
-  list(APPEND BRAYNSIO_LINK_LIBRARIES Brion Brain)
+  list(APPEND BRAYNSIO_LINK_LIBRARIES Brion Brain ${HDF5_LIBRARIES})
 endif()
```
eile: All the above libraries leak through the API?

favreau: Not quite sure if I understand what you mean there?

eile: All link libraries are now declared public, which means that users of BraynsIO have full access to all symbols to these libraries. One only needs to declare link libs public if they are part of the API, e.g., BraynsIO uses some boost type in its API.

PR #36, opened 2016-10-14
Fixed colormap bug in OptiX renderer

favreau:
Using size() on a buffer in an OptiX kernel does not return the number of elements in the buffer. This value has to be specified explicitly in a separate attribute.


favreau:
Retest this please

PR #40, opened 2016-11-01
Use lexis/render/Stream event to Deflect stream available via HTTP

tribal-tec:


plugins/extensions/plugins/DeflectPlugin.cpp
```diff
@@ -187,12 +206,9 @@ void DeflectPlugin::_send(
...
     deflect::ImageWrapper deflectImage( imageData, windowSize.x(), windowSize.y(), deflect::RGBA );
 
-    ApplicationParameters& applicationParameters =
-        _brayns.getParametersManager().getApplicationParameters();
-
-    deflectImage.compressionQuality = applicationParameters.getJpegCompression();
+    deflectImage.compressionQuality = _params.getQuality();
     deflectImage.compressionPolicy =
-        ( deflectImage.compressionQuality != 100 ) ?
+        _params.getCompression() ?
```
favreau: Following our discussion, it's probably better to have 2 different parameters for the quality and the activation of the compression. For the records, compressing to 100% gives the same quality as compression disabled, but removes the Alpha information.

favreau:
Nice functionality! Fix the GitExternal.cmake SHA1 (51a2753 -> a853636) and it should be all good.

plugins/extensions/plugins/DeflectPlugin.cpp
tribal-tec: That's what I have now?! getCompression() is ON/OFF and there is getQuality().

tribal-tec:
retest this please

tribal-tec:
Needs approval for merge

PR #43, opened 2016-11-09
Simulation data must be assigned to all renderers

favreau:


favreau:
Yep, I know, thanks for the reminder :-)

PR #44, opened 2016-11-09
FindOptix.cmake: use OPTIX_ROOT from ENV to find OptiX module

rdumusc:
Configures with:
module load BBP/viz/env/2016.10.28
module load BBP/viz/optix/4.0.0

### DIFF
--- a/CMake/FindOptiX.cmake
+++ b/CMake/FindOptiX.cmake
@@ -32,8 +32,14 @@
 set(OptiX_FOUND FALSE)
 set(OptiX_PATH)
 
-# Our initial guess will be within the SDK.
-set(OptiX_INSTALL_DIR "${CMAKE_SOURCE_DIR}/../OptiX" CACHE PATH "Path to OptiX installed location.")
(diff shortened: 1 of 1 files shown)


rdumusc:
Builds correctly but fails to start with:
```
module load BBP/viz/env/2016.10.28
vglrun bin/braynsViewer --engine optix
[....]
[INFO ] Initializing frame buffer
[DEBUG] [PTX] ./ptx/braynsOptiXCudaPlugin_generated_Camera.cu.ptx
terminate called after throwing an instance of 'optix::Exception'
  what():  File not found (Details: Function "RTresult _rtProgramCreateFromPTXFile(RTcontext, const char*, const char*, RTprogram_api**)" caught exception: ./ptx/braynsOptiXCudaPlugin_generated_Camera.cu.ptx, file:/root/sw/wsapps/raytracing/rtsdk/rel4.0/src/Objects/Program.cpp, line: 113)
Aborted (core dumped)
```

favreau:
The ptx file must exist in a subfolder called ptx, located in the same folder as the braynsViewer application.

PR #46, opened 2016-11-09
 	Allow scenes to be reloaded dynamically through reset event

favreau:


favreau:
Retest this please

plugins/engines/optix/OptiXScene.cpp
```diff
@@ -48,34 +49,116 @@ OptiXScene::OptiXScene(
...
+    Scene::reset();
+
+    // Geometry
+    for( auto geometryInstance: _geometryInstances)
+        geometryInstance->destroy();
+    _geometryInstances.clear();
+
+    if( _geometryGroup )
+        _geometryGroup->destroy();
+    _geometryGroup = nullptr;
```
rdumusc: This function could be made much simpler by wrapping all the destroyable objects in unique_ptrs with a custom deleter that calls destroy()

brayns/common/scene/Scene.cpp
```diff
@@ -46,8 +46,14 @@ Scene::Scene(
 
 Scene::~Scene( )
 {
+    reset();
```
favreau: Not really. It is necessary in the inherited classes, for the various engines, but definitely not required in the Abstract Scene class. Removed.

brayns/common/camera/Camera.cpp
```diff
@@ -51,10 +51,13 @@ struct Camera::Impl
...
         const Vector3f& upVector )
     {
+        const Vector3f& p = getPosition();
+        const Vector3f& t = getTarget();
+        const Vector3f& u = getUpVector();
         setPosition( position );
         setTarget( target );
         setUpVector( upVector );
-        modified = true;
+        modified = ( p != position || t != target || u != upVector );
```
rdumusc: yes, then the problem should be addressed in each of the individual setter, they must do (example code):
```
setPosition( newPos ){
if( newPos == _pos )
    return();
_pos = newPos;
}
```

favreau: Sounds good. Done.

plugins/engines/optix/OptiXScene.cpp
favreau: Tried it but in the end, it was less readable.

PR #47, opened 2016-11-09
Make Optix engine's *.ptx files accessible from both build + install folders

rdumusc:


favreau:
Looking at this PR makes me realize that loading PTX files from disk might not be the best approach. Let me check what OptiX offers to make deployment simpler.

plugins/engines/optix/CMakeLists.txt
```diff
@@ -28,8 +28,8 @@ set(BRAYNSOPTIXENGINEPLUGIN_PUBLIC_HEADERS
 )
 
 set(BRAYNSOPTIXENGINEPLUGIN_LINK_LIBRARIES
-    PUBLIC vmmlib braynsCommon braynsIO
-    PRIVATE braynsCommonEnginePlugin ${OptiX_LIBRARIES}
+  PUBLIC vmmlib braynsCommon braynsIO
+  PRIVATE braynsCommonEnginePlugin Lunchbox ${OptiX_LIBRARIES}
```
tribal-tec: if(TARGET Lunchbox) missing somewhere?

rdumusc: correct - will do if final solution uses Lunchbox, but waiting for @favreau who is looking for an alternative

favreau: Ideally, I would like the PTX code to be stored as a resource of the application. The CUDA_WRAP_SRCS macro allows the creation of OBJ files instead of PTX, but the OptiX API does not seem to make it possible to use this representation. Currently discussion with NVIDIA to see what is possible there.

rdumusc:
replaced by https://github.com/BlueBrain/Brayns/pull/48

PR #53, opened 2016-11-17
Camera event cleanup

favreau:


brayns/common/types.h
```diff
@@ -298,13 +298,13 @@ enum MorphologyLayoutType
     ML_GRID,
 };
 
-enum CameraType
+enum class CameraType
```
tribal-tec: just call it 'Camera'?

favreau: Not in this case, it has to be consistent with the rest of the application code.

plugins/engines/ospray/render/OSPRayCamera.cpp
```diff
@@ -32,13 +32,13 @@ OSPRayCamera::OSPRayCamera( const CameraType cameraType )
     std::string cameraAsString;
```
tribal-tec: Forget it, thought it was the FBS...

brayns/common/types.h
tribal-tec: Forget it, thought it was the FBS...

plugins/engines/ospray/render/OSPRayCamera.cpp
favreau: Ok. thx.

PR #66, opened 2016-11-23
Slower the pan speed, increase wheel speed, make pan and pinch exclusive in DeflectPlugin

tribal-tec:


brayns/common/camera/FlyingModeManipulator.cpp
```diff
@@ -72,8 +72,8 @@ void FlyingModeManipulator::dragRight( const Vector2i& to,
 void FlyingModeManipulator::dragMiddle( const Vector2i& to,
                                         const Vector2i& from )
 {
-    const float x = ( to.x() - from.x( )) * getMotionSpeed();
-    const float y = ( to.y() - from.y( )) * getMotionSpeed();
+    const float x = ( to.x() - from.x( )) * getMotionSpeed() / 100.f;
```
rdumusc: Hmmm, what are those four "getMotionSpeed() / 100.f", can't that be part of the motion speed? Or at least defined as a constant?

PR #76, opened 2016-12-02
Resolved circular dependencies

favreau:


brayns/common/engine/Engine.h
```diff
@@ -0,0 +1,186 @@
...
+{
+
+public:
+
+    Engine( ParametersManager& parametersManager );
+    virtual ~Engine() {}
+
+    /**
+       Returns the name of the engine
+       @return the name of the engine
```
rdumusc: a bit repetitive, keep just the @return

brayns/common/engine/Engine.h
```diff
@@ -0,0 +1,186 @@
...
+ * An engine holds a native implementation of a scene, a camera, a frame buffer
+ * and of one or serveral renderers according to the capatilities of the
+ * acceleration library.
+ */
+class Engine
+{
+
+public:
+
+    Engine( ParametersManager& parametersManager );
```
rdumusc: mark explicit, document the parametersManager argument

brayns/common/engine/Engine.h
```diff
@@ -0,0 +1,186 @@
...
+    FrameBuffer& getFrameBuffer() { return *_frameBuffer; }
+
+    /** Gets the camera */
+    Camera& getCamera() { return *_camera; }
+
+    /** Gets the renderer */
+    Renderer& getRenderer() { return *_renderers[_activeRenderer]; }
+
+    /** Gets the keyboard handler */
+    KeyboardHandler& getKeyboardHandler() { return *_keyboardHandler; }
```
rdumusc: does this have to be part of the engine?

brayns/common/engine/Engine.h
```diff
@@ -0,0 +1,186 @@
...
+
+    /** Sets the active renderer */
+    void setActiveRenderer( const std::string& renderer );
+
+    /** Gets the parameter manager */
+    ParametersManager& getParametersManager() { return _parametersManager; }
+
+    /** Gets the camera manipulator */
+    AbstractManipulator& getCameraManipulator() { return *_cameraManipulator; }
```
rdumusc: does this have to be part of the engine?

brayns/common/engine/Engine.h
```diff
@@ -0,0 +1,186 @@
...
+       @param frameSize New size for the buffers
+
+       @todo Must be removed and held by the render method above
+    */
+    void reshape( const Vector2ui& frameSize );
+
+    /**
+       Register keyboard shorcuts
+    */
+    void registerKeyboardShortcuts();
```
rdumusc: couldn't this be  removed? If not, then more documentation is needed so that users know when to call it.

brayns/common/engine/Engine.cpp
```diff
@@ -0,0 +1,402 @@
...
+
+namespace brayns
+{
+
+Engine::Engine( ParametersManager& parametersManager )
+    : _parametersManager( parametersManager )
+    , _scene( 0 )
+    , _camera( 0 )
+    , _frameBuffer( 0 )
+    , _keyboardHandler( 0 )
```
rdumusc: you can remove the 4 lines above, smart pointer initialize themselves

brayns/common/engine/Engine.h
favreau: Well, I also wandered, but in the end, I think it's good to have it there. Different engines can hold different keys. And I think it's better than in the Brayns class anyway.

brayns/common/engine/Engine.cpp
favreau: Ok

brayns/common/engine/Engine.h
favreau: Same argument here. I'd rather have the engine self contained, rather than having some stuff in the Brayns class and some others in the engine class.

favreau: You're actually totally right, The Brayns class should take care of handling external events (keyboard) just as it already handles the data sources. Fixed! and thanks!

brayns/common/engine/Engine.h
favreau: dito

brayns/common/engine/Engine.h
```diff
@@ -0,0 +1,173 @@
...
+    void _disableShading();
+    void _increaseAmbientOcclusionStrength();
+    void _decreaseAmbientOcclusionStrength();
+    void _resetTimestamp();
+    void _infiniteTimestamp();
+    void _toggleShadows();
+    void _toggleSoftShadows();
+    void _increaseSamplesPerRay();
+    void _decreaseSamplesPerRay();
+    void _toggleLightEmittingMaterials();
```
rdumusc: you can remove all the declarations above

tests/CMakeLists.txt
```diff
@@ -4,7 +4,9 @@
 #
 # This file is part of Brayns <https://github.com/BlueBrain/Brayns>
 
-set(TEST_LIBRARIES ${Boost_UNIT_TEST_FRAMEWORK_LIBRARY} brayns Lunchbox)
+common_find_package(Lunchbox)
```
rdumusc: it should be grouped with the others in the top-level CMakeLists normally

favreau: Ok

brayns/common/engine/Engine.h
favreau: Oops. Thanks

PR #86, opened 2016-12-12
Added simulation histogram

favreau:


plugins/extensions/plugins/ZeroEQPlugin.cpp
```diff
@@ -173,6 +173,10 @@ void ZeroEQPlugin::_setupHTTPServer()
         std::bind( &ZeroEQPlugin::_requestViewport, this ));
     _remoteViewport.registerDeserializedCallback(
         std::bind( &ZeroEQPlugin::_viewportUpdated, this ));
+
+    _httpServer->handle( _remoteHistogram );
```
tribal-tec: handleGET only

favreau: Good point, thanks

PR #91, opened 2016-12-15
Fixed streaming to Deflect

favreau:


plugins/extensions/plugins/DeflectPlugin.cpp
```diff
@@ -71,8 +72,9 @@ namespace brayns
         '*', "Enable/Disable Deflect streaming",
                 [&] { _params.setEnabled( !_params.getEnabled( )); });
 
-    // ensure streaming is off initially, otherwise our logic in run() won't work
-    _params.setEnabled( false );
+    _params.setEnabled( true ); // Streaming will only be activated if
```
tribal-tec: I hope you tested it from HTTP then, because I did put that comment on purpose from what I remember.

favreau: Well yes, I tested all possible configurations:
- Enable/disable stream from HTTP
- Start application with and without DEFLECT_HOST environment variable
- Close stream from Tide
All of them work and the algorithm seems fine to me. Anything else that I should check?

tribal-tec: Ok, good then. Just wanted to be sure :)

PR #102, opened 2017-01-18
simplify finding OSPRay and Embree (Embree found by OSPRay)

jeffamstutz:
This change removes the old (and deprecated) method of finding OSPRay via a FindOSPRay.cmake included in the Brayns source tree. Since OSPRay 0.8.3, we have included a find_package() config as a part of OSPRay's install. Because OSPRay depends on Embree anyway, Embree's headers/libs are also included in OSPRAY_INCLUDE_DIRS/OSPRAY_LIBRARIES. Using this method, you can now rely upon OSPRay's method of finding itself in CMake instead of needing to maintain that yourself.

FYI: Your instructions in the README should be correct for this method. If anything, they should be more robust now in terms of getting Embree/OSPRay setup.

Also FYI: I did not try this on the recommended OS (Ubuntu 16.04), rather on Arch linux. Everything OSPRay related ought to be identical between the two, but that is *technically* a difference.

Last FYI: The spelling of find_package(ospray) is case sensitive, as "ospray" is what the OSPRay project uses for its find_package configs that it installs.


jeffamstutz:
I also forgot to add:

If you need any of OSPRay's CMake macros, you can access them by adding the following to your CMake code before you make any OSPRay CMake macro calls

```
include(${OSPRAY_USE_FILE})
```

This will bring any OSPRAY_* macros as well as ispc target compile macros as well for setting up ISA to compile for and wrappers for creating libs/exes with ispc files in it (like the OSPRay project does).

favreau:
Looking cool! Thanks for that :-) Just to make sure, I am right to say that this requires OSPRay to be previously deployed in the default /usr/local folder first?

favreau:
Retest this please

jeffamstutz:
No, CMake will look wherever the env variable "ospray_DIR" points to (like in the Brayns README). If that is not set, CMake *should* look in standard places, including (but not limited to) /lib, /usr/lib, and /usr/local/lib. However, CMake's default search paths is a CMake thing in general and not specific to OSPRay. The key is that our find_package config that we install follows a specific naming convention with the name of the environment variable and the name of the project you are finding via find_package(). In this case, the spelling of "ospray" is the same for all of them....all true for "embree" too, but we transparently look for that when you find OSPRay anyway.

favreau:
retest this please

favreau:
retest this please

favreau:
Ok, thanks for the explanations. Unfortunately, our CI system now needs a TBB module in order to build Brayns. This will take a few days until I can merge your PR. Working on it...

jeffamstutz:
No problem. The changes are simple enough (at least what I changed) that you could reject the PR and reimplement in a different changeset that encompasses all the issues that come up. Ultimately it's up to you: I made a PR to have changes associated with my suggestions in my initial discussion. :)

favreau:
Retest this please

favreau:
Thanks @jeffamstutz for the explanations and for providing the solution. Welcome to the project! Merging now.

PR #104, opened 2017-01-20
Correct camera model to follow OpenGL semantics; means that z-axis is inverted

tribal-tec:


tribal-tec:
Retest this please

tribal-tec:
Approve!

PR #105, opened 2017-01-20
Make OSPRay optional

eile:


CMakeLists.txt
```diff
@@ -22,21 +22,17 @@ include(Common)
...
+set(OSPRAY_REQUIRED)
 if(NOT ${CMAKE_SOURCE_DIR} STREQUAL ${PROJECT_SOURCE_DIR} OR
         (TRAVIS AND LSB_DISTRIBUTOR_ID MATCHES "Ubuntu"))
-  common_find_package(OSPRay)
-  if(NOT OSPRAY_FOUND)
-      message(WARNING "[Brayns] OSPRay not found, the project will not be built")
-    include(CommonPackageConfig)
-    return()
-  endif()
+  set(OSPRAY_REQUIRED REQUIRED)
```
tribal-tec: what? Why should ospray be required for brayns? it also works with optix and/or livre. Why this behavioral change?

eile: No behavioural change.

Ask @favreau : I got a 5-line warning in stable, this simply unifies the output with other projects.

favreau: The previous implementation already requires OSPRay, which is a mistake. This PR does not change the behaviour, but does not fix the problem of making OSPRay optional. @eile, feel free to make that change, but we can also do that in a dedicated PR.

PR #108, opened 2017-01-26
Implement runtime-switching of engines; allows also to change volume on-the-fly for Livre

tribal-tec:


brayns/Brayns.cpp
```diff
@@ -178,24 +189,21 @@ struct Brayns::Impl
...
+
         const Vector2ui windowSize = _parametersManager->getApplicationParameters().getWindowSize();
+
         _engine->reshape( windowSize );
+        _engine->preRender();
 
+        _extensionPluginFactory->execute( *_engine );
+
+        _engine->reshape( windowSize );
         _engine->preRender();
```
favreau: Maybe check the engine pointer and only apply the reshape + preRender if it was modified.

brayns/common/engine/Engine.h
```diff
@@ -112,7 +112,7 @@ class Engine
     void setDefaultEpsilon();
 
     /**
-       @brief Makes the engine. This means that all attributes, including geometry, material,
+       @brief Makes the engine dirty. This means that all attributes, including geometry, material,
```
favreau: Engine re-creation

brayns/parameters/ParametersManager.h
```diff
@@ -104,6 +104,8 @@ class ParametersManager
      */
     void set( const std::string& key, const std::string& value );
 
+    std::function< void() > recreateEngine;
```
favreau: Maybe you can try to move this function to the Engine class?

plugins/engines/livre/LivreEngine.cpp
```diff
@@ -49,11 +50,34 @@ LivreEngine::LivreEngine( int argc, char **argv,
     // force offscreen rendering
     ::setenv( "EQ_WINDOW_IATTR_HINT_DRAWABLE", "-12" /*FBO*/, 1 /*overwrite*/ );
 
+    // deflect streaming is handled by Brayns, disable Equalizer-integrated stream
+    auto deflectHost = ::getenv( "DEFLECT_HOST" );
```
favreau: Put DEFLECT_HOST in a const var

PR #115, opened 2017-02-06
Fixed transfer function according to new Lexis event

favreau:


plugins/extensions/plugins/ZeroEQPlugin.cpp
```diff
@@ -176,23 +176,15 @@ void ZeroEQPlugin::_setupHTTPServer()
...
         std::bind( &ZeroEQPlugin::_spikesUpdated, this ));
     _remoteSpikes.registerSerializeCallback(
         std::bind( &ZeroEQPlugin::_requestSpikes, this ));
 
-    _httpServer->handle( _remoteColormap );
-    _remoteColormap.registerDeserializedCallback(
-        std::bind( &ZeroEQPlugin::_colormapUpdated, this ));
-    _remoteColormap.registerSerializeCallback(
-        std::bind( &ZeroEQPlugin::_requestColormap, this ));
+    _httpServer->handlePUT( _remoteMaterialLUT );
```
tribal-tec: rename to _remoteColorMaps?

tribal-tec: no support for GET?

plugins/extensions/plugins/ZeroEQPlugin.cpp
```diff
@@ -455,54 +398,51 @@ bool ZeroEQPlugin::_requestLookupTable1D( )
...
-    _remoteColormap.setRange( { range[0], range[1] } );
-    return true;
+
+    auto& contributions = transferFunction.getContributions();
+    for( const auto& contribution: _remoteMaterialLUT.getContributionVector( ))
+        contributions.push_back( contribution );
+
+    Vector2f colormapRange = DEFAULT_TRANSFER_FUNCTION_RANGE;
+    const auto& range = _remoteMaterialLUT.getRange();
+    if( range[0] != range[1] )
```
tribal-tec: why is this necessary?

plugins/extensions/plugins/ZeroEQPlugin.h
```diff
@@ -137,15 +125,10 @@ class ZeroEQPlugin : public ExtensionPlugin
...
 
     /**
-     * @brief This method is called when the colomap is updated by a ZeroEQ event
-     */
-    void _colormapUpdated();
-
-    /**
-     * @brief This method is called when the colormap is requested by a ZeroEQ event
-     * @return True if the method was successful, false otherwise
+     * @brief This method is called when the material lookup table is updated by
```
tribal-tec: colorMaps. No more material lookup table

plugins/extensions/plugins/ZeroEQPlugin.h
```diff
@@ -137,15 +125,10 @@ class ZeroEQPlugin : public ExtensionPlugin
...
+     *        a ZeroEQ event
      */
-    bool _requestColormap();
+    void _materialLUTUpdated();
```
tribal-tec: dito

plugins/extensions/plugins/ZeroEQPlugin.cpp
favreau: _remoteMaterialLUT is more consistent with other class member names I think. And no GET for now since the renderer is not the 'owner' of the colormap.

plugins/extensions/plugins/ZeroEQPlugin.cpp
favreau: It's not, it's a temporary hack. I removed it.

plugins/extensions/plugins/ZeroEQPlugin.cpp
tribal-tec: materialLUT was the old name. The consistency so far was to name the variables close to the class name.

favreau: As discussed, kept only the PUT command, and removed the default values.

PR #121, opened 2017-02-16
Moved performance tests to dedicated folder

favreau:


favreau:
Retest this please

hernando:
If the test is in a directory called perf you don't need to prefix its name by perf.

favreau:
Yes, I remember that, but that still makes sense to start the name of the file with perf.

hernando:
For me it's redundant.

favreau:
Renamed perfBrayns.cpp to renderingSettings.cpp

PR #122, opened 2017-03-01
Fixed bug in CA spheres position management

favreau:


favreau:
Bug introduced in a last minute change in the previous PR... Now fixed :-)

PR #126, opened 2017-03-03
Improve build system by allowing to specify EMBREE and OSPRAY DIRECTO…

adevress:
…RY directly with OSPRAY_ROOT and EMBREE_ROOT


adevress:
ping @favreau  :)

PR #132, opened 2017-03-10
Decouple Brayns dependencies from git_subprojects 

adevress:
Decouple Brayns dependencies from git_subprojects to allow to build Brayns standal
     with external dependencies the option -DDISABLE_SUBPROJECTS=TRUE


adevress:
ping @favreau :)

Last modification before redeployment on Cineca with Opsray 1.20 :)

adevress:
Ok Done, Most of the problems were coming from the fact the file .gitsubproject script staff in Brayns and this file is not read / stripped away when building with -DDISABLE_SUBPROJECTS

PR #135, opened 2017-03-16
Added Global Illumination to OptiX volume rendering

favreau:


plugins/engines/optix/cuda/renderer/Phong.h
```diff
@@ -88,6 +88,9 @@ rtDeclareVariable(uint, colorMapSize, , );
 
 rtBuffer<uchar4, 2> output_buffer;
 
+#define VOLUME_SHADOW_SAMPLES_FACTOR 32
+#define VOLUME_DEFAULT_ALPHA 2.f
```
hernando: Why don't you use const variables? They are fine in headers as they don't violate the single definition rule.

PR #142, opened 2017-04-03
Added simulation model

favreau:
The geometry is defined by the primary model, but the shading is computed according to a second, non-visible model that contains the simulation values. In the case of circuit rendering for examples, neurons are meshes, and compartment reports are cones and cylinders that fill inside of the neuron mesh. When an intersection is found at the surface of a mesh, a ray defined by the intersection point and the inverse of the normal to the surface is launched to hit the secondary model. When the secondary model is intersected, the simulation value is retrieved and used to shade the surface of the mesh.


favreau:
Retest this please

favreau:
Retest this please

PR #148, opened 2017-04-12
Replaced enum by enum class, once and for all

favreau:


brayns/common/types.h
```diff
@@ -202,14 +192,25 @@ typedef std::shared_ptr<DeflectPlugin> DeflectPluginPtr;
...
 /** Define the frame buffer format */
-enum FrameBufferFormat
+enum class FrameBufferFormat
 {
-    FBF_RGBA_NONE,
-    FBF_RGBA_I8,
-    FBF_BGRA_I8,
-    FBF_RGB_I8,
-    FBF_RGBA_F32,
+    rgba_none,
```
rdumusc: what is rgba_none? Is it just none?

brayns/common/types.h
```diff
@@ -303,14 +293,14 @@ const std::string TEXTURE_NAME_SKYBOX = "SKYBOX";
...
-                       // and light emission
-    MT_SHADES_OF_GREY, // 255 shades of grey
-    MT_GRADIENT,       // Gradient from red to yellow
-    MT_PASTEL_COLORS,  // Random pastel colors
+    none,           // Random colors
+    random,         // Random materials including transparency, reflection,
+                    // and light emission
+    shades_of_grey, // 255 shades of grey
+    gradient,       // Gradient from red to yellow
+    pastel,         // Random pastel colors
```
rdumusc: extra , after pastel

brayns/common/types.h
favreau: Do idea, that makes no sense :-( I removed it. Probably a left over from some previous experiments. Thanks.

brayns/common/types.h
favreau: Removed too

PR #149, opened 2017-04-13
Fixed circuit bounding box management

favreau:


brayns/io/MorphologyLoader.cpp
```diff
@@ -995,6 +1030,12 @@ bool MorphologyLoader::importSimulationData(const servus::URI& circuitConfig,
     return true;
 }
 
+bool MorphologyLoader::_positionInCircuitBoundingBox(const Vector3f& position)
```
tribal-tec: const function?

hernando:
Aren't you using brain for accessing the circuit data? I have the feeling that if you did your code would be simpler. At the same time, there's no function there to find out the bounding box of the circuit, but it could and maybe should be added.

favreau:
I am using brain indeed, in the importCircuit methods. What makes you think I am not?

hernando:
True, now that I look again there're some identifiers inside the brain namespace. I just took a quick look but the lines with mv3support? seemed fishy to me. Why do you have to make a distinction?

favreau:
I totally agree with you regarding the mvd3support, I did that because I needed to get the layer, m-type, e-type, etc of the neuron and as far as I understand, this is currently only supported in brion. If there is a way to get this information in brain, I am happy to modify the code.

favreau:
Ping

PR #160, opened 2017-04-28
Fixed OpenMP issue when loading morphologies in parallel

favreau:


hernando:
It's quite hard to understand what was the problem and what's the solution. Can you add some comments in the commit message?

favreau:
Basically, the 'private' vectors of geometry (Spheres, Cylinders and Cones) used to load individual morphologies where not defined in the right scope, meaning that they were accessed by several thread at the same time. Since std::vectors are not threadsafe, this was causing an exception. The purpose of this fix is to declare those private vectors in the right scope to make sure that they are accessed by one single thread.

hernando:
My request was to write that as part of the commit message, which is very terse. I know I repeat myself a lot with that, bu the commit message is the only thing you can be sure that will remain for the record.

favreau:
Fair enough, and thanks for the clarification. I added it to the commit message.

timocafe:
my 2 cents, TBB provides concurrent vector with stl API.

https://software.intel.com/en-us/node/506203

As you are already using TBB and avoid mixup TBB/OPENMP

PR #166, opened 2017-05-05
Various fixes found during NESTLoader & cache file debugging

tribal-tec:
- speedup loading by using boost::progress_display
- fix JPEG encoded image (wrong pixelformat)
- fix off-by-one error in histogram calculation
- abort if cache writing gets bad, e.g. disk full

### DIFF
--- a/brayns/Brayns.cpp
+++ b/brayns/Brayns.cpp
@@ -404,13 +404,11 @@ private:
         // Load PDB File
         auto& geometryParameters = _parametersManager->getGeometryParameters();
         const std::string& folder = geometryParameters.getPDBFolder();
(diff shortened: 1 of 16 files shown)


brayns/Brayns.cpp
```diff
@@ -509,16 +507,32 @@ struct Brayns::Impl
         {
             size_t nbMaterials;
             NESTLoader loader(geometryParameters);
+
             loader.importCircuit(circuit, scene, nbMaterials);
```
favreau: The circuit should only be loaded if Brayns can attach to the cache file. Otherwise, it's a waste of time. Also, the loading should fail is no cache file is specified.

PR #168, opened 2017-05-11
Fixed material management

favreau:
Until now, a fixed number of materials was initialized at application start-up. The whole material management has been revisited to allow a dynamic number of elements.


brayns/common/types.h
```diff
@@ -284,15 +283,15 @@ struct Histogram
 /** Some 'special' materials are used by Brayns to acomplish specific features
  *  such as skyboxes.
  */
+const size_t DEFAULT_NB_MATERIALS = 200;
```
tribal-tec: rm

plugins/engines/ospray/OSPRayScene.cpp
```diff
@@ -989,69 +994,84 @@ void OSPRayScene::commitLights()
 
 void OSPRayScene::commitMaterials(const bool updateOnly)
 {
-    for (const auto& renderer : _renderers)
+    _ospMaterialData = 0;
+
+    // Determine how many materials need to be created
+    size_t maxId = -std::numeric_limits<size_t>::max();
```
tribal-tec: shouldn't be min? -max for unsigned overflows to 0 or 1, no?

favreau: hum... yes, 0, definitely :-) Thanks

PR #174, opened 2017-06-06
Activate optional dependencies for continuous integration

favreau:


CMakeLists.txt
```diff
@@ -17,19 +17,26 @@ endif()
...
 
-option(BRAYNS_BRION_ENABLED "Activate Brion" OFF)
-option(BRAYNS_DEFLECT_ENABLED "Activate streaming to display wall" OFF)
-option(BRAYNS_NETWORKING_ENABLED "Activate networking interfaces" OFF)
-option(BRAYNS_ASSIMP_ENABLED "Activate assimp library for standard 3D file format management" OFF)
-option(BRAYNS_IMAGEMAGICK_ENABLED "Activate IMAGEMAGICK library for standard image file format management" OFF)
+include(Common)
+
+# By default, optional dependencies are disabled
+set(BRAYNS_DEPENDENCY_ENABLED OFF)
```
eile: As a user, this is hard to understand (which is why it needs a comment). BRAYNS_USE_OPTIONAL_DEPENDENCIES?

eile: Why is it not auto, aka on-if-found!?

favreau: We discussed the auto detection of libraries and this is something I agree this should be done. I'll do that in a later PR though. Thanks.

favreau:
Retest this please

CMakeLists.txt
```diff
@@ -96,6 +100,8 @@ if(BRAYNS_BENCHMARK_ENABLED)
   add_subdirectory(apps/BraynsBenchmark)
 endif()
 
+common_find_package_post()
```
tribal-tec: needs to be moved up before any add_subdirectory

PR #175, opened 2017-06-06
Fix for livre change

tribal-tec:


tribal-tec:
needed after https://github.com/BlueBrain/Livre/pull/397

PR #176, opened 2017-06-16
Fixed loading issues with morphology meshes

favreau:


favreau:
Hum, it's basically a set of small fixes. I am not sure if this should be added to the change log. Important thing is that it's now fixed, and nobody raised an issue so far anyway regarding those specific fixes.

PR #195, opened 2017-08-10
Add support for DeflectPixelOp for optimized streaming with OSPRay

tribal-tec:


plugins/engines/ospray/OSPRayEngine.cpp
```diff
@@ -46,16 +48,17 @@ OSPRayEngine::OSPRayEngine(int argc, const char** argv,
                     << std::endl;
     }
 
+    RenderingParameters& rp = _parametersManager.getRenderingParameters();
+    if (!rp.getModule().empty())
+        ospLoadModule(rp.getModule().c_str());
```
favreau: Maybe ospLoadModule returns a value that would be a better check for knowing if the module is loaded?

plugins/extensions/plugins/DeflectPlugin.cpp
```diff
@@ -102,12 +108,30 @@ bool DeflectPlugin::run(Engine& engine)
...
+    {
+        if (_initializeDeflect(observerOnly))
+        {
+            const auto& windowSize = engine.getParametersManager()
+                                         .getApplicationParameters()
+                                         .getWindowSize();
+            deflect::SizeHints sizeHints;
+            sizeHints.preferredWidth = windowSize.x();
+            sizeHints.preferredHeight = windowSize.y();
+            sizeHints.minWidth = 64;
```
favreau: Why 64? Magic value? or a parameter depending  on the engine? embree tile size? Deflect tile size?

PR #196, opened 2017-08-16
Fix compartment report mapping

tribal-tec:


brayns/parameters/GeometryParameters.h
```diff
@@ -261,7 +261,7 @@ class GeometryParameters : public AbstractParameters
     std::string _circuitReport;
     double _circuitStartSimulationTime;
     double _circuitEndSimulationTime;
-    double _circuitSimulationStep;
+    double _circuitSimulationStep{0};
```
favreau: Why is this required?

tribal-tec: because it's uninitialized?

PR #202, opened 2017-09-05
Added CircuitConfigurationBuilder Rest API

favreau:


plugins/extensions/plugins/ZeroEQPlugin.cpp
```diff
@@ -20,6 +20,8 @@
 
 #include "ZeroEQPlugin.h"
 
+#include <ZeroEQ/zeroeq/http/helpers.h>
```
rdumusc: won't work if ZeroEQ package is in an installed location.
#include <zeroeq/http/helpers.h>

plugins/extensions/plugins/ZeroEQPlugin.cpp
```diff
@@ -178,6 +182,32 @@ void ZeroEQPlugin::handle(servus::Serializable& object)
...
+            ::zeroeq::http::Code::SERVICE_UNAVAILABLE);
+
+    blueConfig << "Run Default" << std::endl << "{" << std::endl;
+    for (auto param : params)
+    {
+        std::replace(param.begin(), param.end(), '=', ' ');
+        blueConfig << param << std::endl;
+    }
+    blueConfig << "}" << std::endl;
+    blueConfig.close();
```
rdumusc: make a second function:
bool _writeBlueConfigFile(const std::string& filename, const strings& params);

plugins/extensions/plugins/ZeroEQPlugin.cpp
```diff
@@ -178,6 +182,32 @@ void ZeroEQPlugin::handle(servus::Serializable& object)
...
+    boost::split(params, paramsAsString, boost::is_any_of("&"));
+
+    const std::string blueConfigFilename = "/tmp/BlueConfig";
+    std::ofstream blueConfig(blueConfigFilename);
+    if (!blueConfig.good())
+        ::zeroeq::http::make_ready_response(
+            ::zeroeq::http::Code::SERVICE_UNAVAILABLE);
+
+    blueConfig << "Run Default" << std::endl << "{" << std::endl;
+    for (auto param : params)
```
rdumusc: const auto& param

plugins/extensions/plugins/ZeroEQPlugin.cpp
favreau: Done

hernando:
I don't know if this is a can of worms that we want to open now, but I think that CircuitConfig are already being registered in the neuroinformatics platform with unique IDs and that you can get either the path or the text from it. I'd make sense to track what's the progress there and consider adding support for this in Brion or viztools

hernando:
I'm saying this because (without knowing the precise details) this commit doesn't look to be in the right place.

PR #204, opened 2017-09-13
Fixed access to simulation data

favreau:
This is a temporary fix. The opening of the circuit within Brayns needs to be refactored so that the code appears only once in the source.

### DIFF
--- a/brayns/Brayns.cpp
+++ b/brayns/Brayns.cpp
@@ -38,7 +38,6 @@
 #include <brayns/io/ProteinLoader.h>
 #include <brayns/io/TransferFunctionLoader.h>
 #include <brayns/io/XYZBLoader.h>
-#include <brayns/io/simulation/CircuitSimulationHandler.h>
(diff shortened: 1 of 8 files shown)


favreau:
@hernando, I know, I was not sure actually. I don't think this will be lost, but the whole data access in Brayns is a larger discussion that will determine the final implementation. Typically, data should be loaded in a asynchronous way and the connection to the data source should be kept alive for as long as Brayns is running. Implementing all this properly is more work than it seems and not in the current priorities, so I would stick to the hot fix for now, but I am happy do discuss a better hot fix if you want?

tribal-tec:
What about doing _loadCompartmentReport() as part of _loadCircuitConfiguration() in Brayns::_loadData()? This way you can pass the GIDs from the circuit loading to the compartment report initialization. Unless you see a point in loading compartment reports w/o loading a circuit, but I think such case does not make sense.

favreau:
@tribal-tec Good idea, let me give it a try!

PR #214, opened 2017-09-22
Use animation frame uint32_t instead of timestamp float

tribal-tec:


brayns/common/vocabulary/parameters.fbs
```diff
@@ -146,7 +146,7 @@ enum Engine: string {
 
 table Settings {
     engine: Engine;
-    timestamp: float;
+    animation_frame: uint;
```
favreau: I think this one should be removed. It duplicates the frame event.

tribal-tec: Good point. Done

hernando:
Just curious. Why changing this to use frame numbers when the compartment report reader uses time stamps and spike datasets have no frames at all? Additionally, I really think that frame numbers is not what users should be seeing, is this change becoming visible in the public API?

tribal-tec:
The animation frame is used for other things as well, like volume time series, morphology growing etc. It was only used so far as a frame, like increasing/decreasing by 1 only, no fractions involved. The only place where it actually needs timestamps is the compartment report. No spikes report in Brayns. The API already exposed the frame, the timestamp was redundant and wrong as it represented the frame number as well.

PR #217, opened 2017-09-25
Made it possible to load multiple targets from a circuit

favreau:


brayns/io/MorphologyLoader.cpp
```diff
@@ -111,78 +115,115 @@ class MorphologyLoader::Impl
...
+                }
 
-        if (gids.empty())
-        {
-            BRAYNS_ERROR << "Circuit does not contain any cells" << std::endl;
-            return false;
-        }
+                if (gids.empty())
+                {
+                    BRAYNS_ERROR << "Target does not contain any cells"
```
tribal-tec: print the target name

PR #219, opened 2017-09-26
Fixed empty frames blinking while playbacking simulation in async mode

tribal-tec:
The buffer is shared with OSPRay, so under no circumstances this shall
be reset() unless we assign a new buffer.


tribal-tec:
Retest this please

tribal-tec:
Retest this please

PR #220, opened 2017-09-29
Added neuron connectivity matrix visualizer

favreau:


brayns/Brayns.cpp
```diff
@@ -85,6 +86,10 @@ struct Brayns::Impl
         _parametersManager->parse(argc, argv);
         _parametersManager->print();
 
+        // Initialize Mesh loader
+        _meshLoader.reset(
```
tribal-tec: why pointer? can we not have a local meshloader everytime we need one? Does it have to be a permanent member of Brayns.cpp?

brayns/Brayns.cpp
```diff
@@ -521,6 +528,21 @@ struct Brayns::Impl
 
 #if (BRAYNS_USE_BRION)
     /**
+        Loads data from a neuron matrix file (command line parameter
+       --matrix-file)
```
tribal-tec: --connectivity-file rather than matrix? Sounds too generic imo. In the end, everything is a matrix :)

brayns/io/MatrixLoader.cpp
```diff
@@ -0,0 +1,276 @@
+/* Copyright (c) 2015-2017, EPFL/Blue Brain Project
+ * All rights reserved. Do not distribute without permission.
+ * Responsible Author: Jafet Villafranca Diaz <jafet.villafrancadiaz@epfl.ch>
```
tribal-tec: jafet? really?

brayns/io/MatrixLoader.cpp
```diff
@@ -0,0 +1,276 @@
...
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include <H5Cpp.h>
```
tribal-tec: h5cpp is deprecated, use highfive instead. Or coordinate with Mohammed and Juan to get this into brion, where highfive is already used.

brayns/io/MeshLoader.cpp
```diff
@@ -322,4 +322,20 @@ void MeshLoader::_createMaterials(Scene& scene, const aiScene* aiScene,
...
     }
 }
+
+std::string MeshLoader::getMeshFilenameFromGID(const uint64_t gid)
+{
+    const auto meshedMorphologiesFolder =
+        _geometryParameters.getCircuitMeshFolder();
+    auto meshFilenamePattern =
+        _geometryParameters.getCircuitMeshFilenamePattern();
+    std::string gidAsString = std::to_string(gid);
```
tribal-tec: const

brayns/io/MeshLoader.h
```diff
@@ -55,7 +55,6 @@ class MeshLoader
      * @return true if the file was successfully imported. False otherwise.
      */
     bool importMeshFromFile(const std::string& filename, Scene& scene,
```
tribal-tec: doxygen update

brayns/parameters/GeometryParameters.h
```diff
@@ -241,6 +241,16 @@ class GeometryParameters : public AbstractParameters
      * Return the full path of the file containing a scene description
      */
     std::string getSceneFile() const { return _sceneFile; };
+    /** File containing neuron matrix */
+    std::string getMatrixFile() const { return _matrixFile; }
```
tribal-tec: const ref for all the returns here please

brayns/Brayns.cpp
favreau: We need a global meshloader indeed (and it's a pain). The reason behind it is that triangle faces use indices to vertices. When you load several meshes from different files, indices have to remain consistent. We could have many mesh loaders, but they would still have to aggregate at some point, so using a single one is the simple solution.

brayns/io/MatrixLoader.cpp
favreau: Yep, he initiated the work. I added me, but I am not removing him ;)

brayns/io/MeshLoader.cpp
favreau: Done

brayns/io/MeshLoader.h
favreau: Done

brayns/Brayns.cpp
tribal-tec: the pointer is still not necessary. using proper initialization order in ctor should suffice

brayns/io/MeshLoader.h
```diff
@@ -39,7 +39,7 @@ namespace brayns
 class MeshLoader
 {
 public:
-    MeshLoader(GeometryParameters& geometryParameters);
+    MeshLoader();
```
tribal-tec: why removed? the geometry params are always the same, so giving them once in the construction is fine.

favreau: MeshLoader is a class member of Brayns and we do not have the geometry parameters yet when the MeshLoader is created. I could add a setGeometryParameters to the MeshLoader class but this would not make much difference with the current implementation. What do you think?

brayns/parameters/GeometryParameters.h
```diff
@@ -241,6 +241,22 @@ class GeometryParameters : public AbstractParameters
      * Return the full path of the file containing a scene description
      */
     std::string getSceneFile() const { return _sceneFile; };
+    /** File containing neuron matrix */
+    std::string getConnectivityFile() const { return _connectivityFile; }
```
tribal-tec: still no const refs here

Issue #200, opened 2017-09-01
Improve CMake options handling


rdumusc:
- Move all the common_find_package() calls to the top-level CMakeLists. This is needed so that the return() from the common_find_package macro works correctly if a REQUIRED dependency is not found when Brayns is a subproject.
- Replace all
```
option(BRAYNS_ENABLE_FOO "Enable Foo support" OFF)
if(BRAYNS_ENABLE_FOO)
  common_find_package(foo)
endif()
```
with:
```
common_find_package(foo)
if(FOO_FOUND)
  option(BRAYNS_ENABLE_FOO "Enable Foo support" ON)
endif()
```
This logic has the advantage of enabling all the options that are found by default and still lets users disable them if need be.

PR #221, opened 2017-10-02
Simplify CMake dependency handling and fix OSX build

rdumusc:
See #200 for details, I wasn't sure about optional .gitsubprojects so I left them as-is for now.


favreau:
I wanted to this ages ago. Thanks for doing it! Really appreciated.

PR #229, opened 2017-10-18
Cache file fixes

tribal-tec:
* Optimize reading speed, total loading time went from ~140 to ~60 seconds
* Fix missing progress update
* Don't build default scene when loading cache file

### DIFF
--- a/plugins/engines/ospray/OSPRayScene.cpp
+++ b/plugins/engines/ospray/OSPRayScene.cpp
@@ -369,10 +369,11 @@ void OSPRayScene::_saveCacheFile()
     BRAYNS_INFO << "Scene successfully saved" << std::endl;
 }
 
-void OSPRayScene::_loadCacheFile()
+void OSPRayScene::loadSceneFromCacheFile()
 {
-    const std::string& filename =
-        _parametersManager.getGeometryParameters().getLoadCacheFile();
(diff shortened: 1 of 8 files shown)


plugins/engines/ospray/OSPRayScene.cpp
```diff
@@ -921,7 +906,7 @@ void OSPRayScene::buildGeometry()
     BRAYNS_INFO << "---------------------------------------------------"
                 << std::endl;
 
-    if (!_parametersManager.getGeometryParameters().getSaveCacheFile().empty())
+    if (!geomParams.getSaveCacheFile().empty())
```
rdumusc: this should be outside the buildGeometry() function

plugins/engines/ospray/OSPRayScene.h
```diff
@@ -91,7 +94,6 @@ class OSPRayScene : public brayns::Scene
     uint64_t _serializeCones(const size_t materialId);
     uint64_t _buildMeshOSPGeometry(const size_t materialId);
 
-    void _loadCacheFile();
     void _saveCacheFile();
```
rdumusc: if saveSceneToCacheFile is empty, move contents to saveSceneToCacheFile()

PR #231, opened 2017-10-20
Minor cleanups, remove Engine::getParametersManager()

rdumusc:


brayns/Brayns.cpp
```diff
@@ -91,35 +90,28 @@ struct Brayns::Impl
...
-        _keyboardHandler.reset(new KeyboardHandler());
         _registerKeyboardShortcuts();
 
-        // Get rendering engine
-        _engineFactory.reset(
-            new EngineFactory(argc, argv, *_parametersManager));
         createEngine();
 
 #if (BRAYNS_USE_DEFLECT || BRAYNS_USE_NETWORKING)
+        // after createEngine() to execue in parallel to scene loading
```
tribal-tec: typo 'execue'

brayns/Brayns.cpp
```diff
@@ -91,35 +90,28 @@ struct Brayns::Impl
...
-        {
-            _engineFactory->remove(_engine);
-            _engine.reset();
-        }
-
         const auto& engineName =
-            _parametersManager->getRenderingParameters().getEngine();
-        _engine = _engineFactory->get(engineName);
+            _parametersManager.getRenderingParameters().getEngine();
+        _engine = _engineFactory.create(engineName);
```
tribal-tec: I would still _engine.reset() explicitly, otherwise two engines are active until the assignment to the unique_ptr is done

plugins/engines/EngineFactory.cpp
```diff
@@ -33,70 +33,46 @@
...
-        {
-            const char** argv = new const char*[_arguments.size()];
-            for (size_t i = 0; i < _arguments.size(); ++i)
-                argv[i] = _arguments[i].c_str();
-            _engines[name] = EnginePtr(
-                new OSPRayEngine(_arguments.size(), argv, _parametersManager));
-            delete[] argv;
-            return _engines[name];
-        }
+            return make_unique<OSPRayEngine>(_argc, _argv, _parametersManager);
```
tribal-tec: iirc, the reason for the copy of argv is that engines might mutate it, aka if the engine is not taking a const char*. Please check that, or we have a look together.

rdumusc: I think that only the LivreEngine had a constness problem, but it is already making a copy of the args internally... I seriously hope that no engine actually alters the CL parameters though!

rdumusc:
argggggg!!

timocafe:
AAAAAARRRRRRRRRRHHHHHHHH

PR #238, opened 2017-10-31
Enh/docker

tristan0x:
On my way to make `braynsViewer` works within containers, here is a small contribution to shrink both Docker image size and build time.

Image size is now 437MB instead of >1GB

The patch leverages _multi-stage builds_, a feature released in Docker 17.05.
More information available in the [official user guide](https://docs.docker.com/engine/userguide/eng-image/multistage-build/#use-multi-stage-builds)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -98,14 +98,15 @@ RUN cksum ${BRAYNS_SRC}/.gitsubprojects && \
     freeglut3-dev \
     libassimp-dev \
     libboost-all-dev \
+    libglew-dev \
     libglu1-mesa-dev \
     libhdf5-serial-dev \
     libjpeg-turbo8-dev \
     libmagick++-dev \
     libturbojpeg \
     libxi-dev \
-    libzmq3-dev \
     libxmu-dev \
+    libzmq3-dev \
     python-pyparsing \
     qtbase5-dev \
     qtdeclarative5-dev \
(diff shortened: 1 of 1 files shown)


favreau:
"On my way" means that your are still working on it, on not ready for review yet?
In any case, this is super cool! Many thanks for your contribution! :)

Dockerfile
```diff
@@ -90,48 +79,80 @@ ADD . ${BRAYNS_SRC}
...
+# 2. create a new image layer containing the
+#    /app/dist directory of this new container
+#    Equivalent to the `docker copy` command.
+COPY --from=builder /app/dist /app/dist
 
 # Add binaries from dist to the PATH
-ENV LD_LIBRARY_PATH $LD_LIBRARY_PATH:${DIST_PATH}/lib
-ENV PATH $PATH:${DIST_PATH}/bin
-
+ENV LD_LIBRARY_PATH $LD_LIBRARY_PATH:/app/dist/lib
```
tribal-tec: why not reusing DIST_PATH instead of hard-coding /app/dist/lib?

rolandjitsu: I agree, just use `DIST_PATH`.

rolandjitsu:
@tristan0x looks great, I was planning on using multi step builds, but never got the time to do it.

tristan0x:
@favreau Regarding BraynViewer, it is still [work in progress](https://github.com/tristan0x/Brayns/wiki).

Dockerfile
tristan0x: 1. You are right, it should not be hardcoded. I will make the required changes.
1. Besides `ENV` and `ARG` are not available in other build stages. they must be redefined.
1. I think we should use `ARG` instead of `ENV` as `DIST_PATH` does not seem to be a required variable in the container. Anybody against?

tristan0x:
Hello @rolandjitsu
Does the last change suits you?

PR #239, opened 2017-10-31
Allow use of OSPRay default renderers

favreau:
And introduces a basic fast BasicRenderer instead of the ExtendedOBJRenderer


brayns/common/material/Material.h
```diff
@@ -66,7 +66,9 @@ class Material
     BRAYNS_API TextureTypes& getTextures() { return _textures; }
     BRAYNS_API void setTexture(const TextureType& type,
                                const std::string& filename);
-
+    BRAYNS_API void lock() { _locked = true; }
+    BRAYNS_API void unlock() { _locked = false; }
+    BRAYNS_API bool locked() { return _locked; }
```
tribal-tec: const function

brayns/common/material/Material.h
```diff
@@ -66,7 +66,9 @@ class Material
     BRAYNS_API TextureTypes& getTextures() { return _textures; }
     BRAYNS_API void setTexture(const TextureType& type,
                                const std::string& filename);
-
+    BRAYNS_API void lock() { _locked = true; }
+    BRAYNS_API void unlock() { _locked = false; }
+    BRAYNS_API bool locked() const { return _locked; }
```
tribal-tec: please put a doxygen here 'cause I already forgot what this was for again

plugins/engines/ospray/OSPRayScene.cpp
```diff
@@ -991,64 +996,60 @@ uint64_t OSPRayScene::_buildMeshOSPGeometry(const size_t materialId)
 
 void OSPRayScene::commitLights()
 {
-    for (auto renderer : _renderers)
+    size_t lightCount = 0;
```
tribal-tec: this is never incremented??

favreau: Fixed

brayns/common/material/Material.h
favreau: Doc added

favreau:
Retest this please

PR #240, opened 2017-11-03
Add websockets support

tribal-tec:
* All PUT messages are supported using a JSON text message with an "event"
  and "data" field where "event" corresponds to the REST endpoint and
  "data" as specified by endpoint/schema. For the camera, it looks like this:
  {
     "event" : "v1/camera",
     "data" : {
       "look_at" : [ 1.0 0.5 0.5 ]
     }
  }
* So far only the camera gets broadcasted as a JSON text message when it
  was modified. The JSON format is the same as for PUT, but contains all camera
  values.

### DIFF
--- a/plugins/extensions/plugins/RocketsPlugin.cpp
+++ b/plugins/extensions/plugins/RocketsPlugin.cpp
@@ -36,6 +36,9 @@
 
 #include <fstream>
 
+#include "json.hpp"
+using json = nlohmann::json;
+
 namespace
 {
 const std::string ENDPOINT_API_VERSION = "v1/";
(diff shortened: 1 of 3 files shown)


rolandjitsu:
@tribal-tec so we have but one shared socket for all events?

rolandjitsu:
I think there might be an issue with the `{data}`. By the looks of it, there seems to be double JSON encoding, on the main socket message and the `{data}` property:
```json
{
    "data": "{\n   \"aperture\" : 0.0,\n   \"eye_separation\" : 0.063500002026557922,\n   \"field_of_view\" : 45.0,\n   \"focal_length\" : 0.0,\n   \"look_at\" : [ 0.73252975940704346, 1.1178562641143799, 1.0505576133728027 ],\n   \"origin\" : [ 2.442774772644043, 4.600367546081543, 7.2228507995605469 ],\n   \"stereo_mode\" : \"none\",\n   \"up\" : [ -0.45402175188064575, -0.82591044902801514, -0.33421853184700012 ]\n}\n",
    "event": "camera"
}
```

@tribal-tec

tribal-tec:
True, good catch. Will provide a fix

PR #241, opened 2017-11-08
Support image and progress over websockets, minor fixes

tribal-tec:


brayns/parameters/ApplicationParameters.h
```diff
@@ -126,6 +133,7 @@ class ApplicationParameters : public AbstractParameters
     std::string _streamId;
     std::string _tmpFolder;
     bool _synchronousMode{false};
+    size_t _imageStreamFPS{60};
```
rdumusc: we are become ambitious now :-)

plugins/extensions/plugins/RocketsPlugin.cpp
```diff
@@ -176,10 +192,33 @@ void RocketsPlugin::_broadcastWebsocketMessages()
...
+
+    if (_engine->getModified())
+    {
+        _requestProgress();
+        const auto& message =
+            _buildJsonMessage(ENDPOINT_PROGRESS, _remoteProgress.toJSON());
+        _httpServer->broadcastText(message);
+    }
+
+    if (_engine->getRenderer().haveNewImage())
```
rdumusc: hmmm, shouldn't this method be called hasNewImage() instead of haveNewImage()?

brayns/parameters/ApplicationParameters.h
tribal-tec: It's just an upper limit. We will never render that fast anyways :)

plugins/extensions/plugins/RocketsPlugin.cpp
tribal-tec: Hmm, true. I'll have to recheck also if this works with Optix.

plugins/extensions/plugins/RocketsPlugin.h
```diff
@@ -179,6 +194,24 @@ class RocketsPlugin : public ExtensionPlugin
     ::brayns::v1::Scene _remoteScene;
     ::brayns::v1::ForceRendering _remoteForceRendering;
     ::brayns::v1::CircuitConfigurationBuilder _remoteCircuitConfigBuilder;
+
+    class Timer
```
rolandjitsu: Can I ask what's the purpose of this timer?

tribal-tec: To limit the image stream to the configurable image stream FPS. 60 FPS by default

PR #243, opened 2017-11-11
Fixed bug in connectivity loader

favreau:


timocafe:
Japan = vacances

Envoyé de mon GameBoy

PR #255, opened 2017-12-11
Add pick() to Renderer; use shift+click in braynsViewer for camera target change

tribal-tec:


hernando:
The camera target setter will be a very useful feature in fact.

PR #260, opened 2017-12-14
Made Brayns Ceph-ready

favreau:


favreau:
Compilation failed with the previous CMake/common sha1. So I fell back to this one. I'll need assistance on Brion to move to the latest.

PR #261, opened 2017-12-18
Moved cache file management to Scene class

favreau:


brayns/common/geometry/Cone.h
```diff
@@ -38,6 +38,7 @@ struct Cone
     {
     }
 
+    Cone() {}
```
tribal-tec: = default

brayns/common/scene/Scene.cpp
```diff
@@ -669,4 +675,325 @@ void Scene::setCylinder(const size_t materialId, const uint64_t index,
...
+    BRAYNS_INFO << "Saving scene to binary file: " << filename << std::endl;
+    std::ofstream file(filename, std::ios::out | std::ios::binary);
+    if (!file.good())
+    {
+        BRAYNS_ERROR << "Could not open cache file " << filename << std::endl;
+        return;
+    }
+
+    const size_t version = CACHE_VERSION;
+    file.write((char*)&version, sizeof(size_t));
```
tribal-tec: why not file << version? everywhere in this file?

brayns/common/scene/Scene.cpp
```diff
@@ -669,4 +675,325 @@ void Scene::setCylinder(const size_t materialId, const uint64_t index,
...
+    std::ifstream file(filename, std::ios::in | std::ios::binary);
+    if (!file.good())
+    {
+        BRAYNS_ERROR << "Could not open cache file " << filename << std::endl;
+        return;
+    }
+
+    // File version
+    size_t version;
+    file.read((char*)&version, sizeof(size_t));
```
tribal-tec: also works backwards :) file >> version;

brayns/common/scene/Scene.cpp
favreau: Not all elements have this operator implemented. Some of those elements being in process of being replaced, I will leave your comments to a later PR if that's acceptable?

brayns/common/scene/Scene.cpp
favreau: dito

PR #264, opened 2017-12-20
Reworked material management

favreau:


tribal-tec:
your change shouldn't affect the visual appearance, so why the testdata has to be updated? Or it was broken before?

tribal-tec:
The update of CMake/common is necessary? It's the latest version?

brayns/common/engine/Engine.h
```diff
@@ -98,15 +98,16 @@ class Engine
 
     /**
        Initializes materials for the current scene
-       @param materialType Predefined sets of colors
+       @param colorMap Predefined color map
              MT_DEFAULT: Random colors
```
tribal-tec: those MT_* values are no longer correct, no?

brayns/common/scene/Scene.cpp
```diff
@@ -247,78 +235,78 @@ void Scene::buildDefault()
...
-        m.setColor(colors[material]);
-        m.setSpecularColor(WHITE);
-        m.setSpecularExponent(10.f);
-        m.setReflectionIndex(material == 4 ? 0.8f : 0.f);
-        m.setOpacity(1.f);
-        auto& trianglesMesh = _trianglesMeshes[NB_SYSTEM_MATERIALS + material];
-        for (size_t i = 0; i < 6; ++i)
+        // Cornell box
+        Material material;
+        material.setColor(colors[i]);
```
tribal-tec: constructor for Material to take some/all of this values?

brayns/common/scene/Scene.cpp
```diff
@@ -699,34 +699,32 @@ void Scene::saveToCacheFile()
...
         file.write((char*)&value, sizeof(float));
-        const bool boolean = material.second.getCastSimulationData();
+        const bool boolean = material.getCastSimulationData();
         file.write((char*)&boolean, sizeof(bool));
         // TODO: Textures
     }
 
     // Save geometry
-    for (auto& material : _materials)
+    ;
```
tribal-tec: rm

brayns/common/scene/Scene.h
```diff
@@ -360,7 +327,56 @@ class Scene
...
+    BRAYNS_API void setMaterial(const size_t index, const Material& material);
+
+    /**
+        Resets materials and builds the system ones (Bounding box, skymap, etc)
+    */
+    BRAYNS_API void resetMaterials();
+
+    /**
+        Commit materials to renderers
+        @param updateOnly If true, materials are not recreated and textures are
```
tribal-tec: wrong doc

brayns/io/MeshLoader.cpp
```diff
@@ -33,6 +33,9 @@
 
 #include <brayns/common/scene/Scene.h>
 
+#define BRAYNS_USE_ASSIMP \
+    1 //************* REMOVE BEFORE PUSH ******************
```
tribal-tec: ??

brayns/common/engine/Engine.h
favreau: Doc is now in the types.h file. Removed.

brayns/common/scene/Scene.cpp
favreau: Done

brayns/common/scene/Scene.h
favreau: Done

brayns/io/MeshLoader.cpp
favreau: Comment was obliviously not big enough... :see_no_evil:

brayns/common/scene/Scene.cpp
favreau: hum... it's either all or none, some would be a bit weird to me. And having them all makes it a long list of arguments. Maybe another idea would to be to define some presets, like Metal, Glass, etc?

favreau:
Regarding the new test file, the current one is broken and does now process material. This PR fixes the problem.

PR #273, opened 2018-01-13
Fixed cache file loading

favreau:


plugins/engines/ospray/OSPRayScene.cpp
```diff
@@ -156,7 +156,10 @@ void OSPRayScene::unload()
 void OSPRayScene::commit()
 {
     if (_model)
+    {
+        BRAYNS_INFO << "Committing main model" << std::endl;
```
tribal-tec: really needed?

favreau: Not vital but I found it convenient to have it while understanding what was going wrong. And it's also more consistent with the simulation model that also traces its commit phase. But I have no strong opinion on that. I can remove it if you think it's useless.

PR #281, opened 2018-01-23
Resize fixes, remove engine recreation

tribal-tec:


favreau:
:'( One of the last remains of the ISC demo is fading away... Fossilized forever in the Github repository.

PR #303, opened 2018-02-05
Move to C++14

tribal-tec:


tribal-tec:
It officially just exists now :) Other unique_ptr are pimpls, and they are just new'd.

PR #331, opened 2018-02-26
Logic for adding Topology Viewer as a usecase plugin

karjonas:
If CMake finds a folder named TopologyViewer in the usecases directory it will include the plugin. There is also some logic for adding custom rpc callbacks to Rockets plugin. This makes it possible to do custom communication between jupyter and the brayns plugin.


plugins/extensions/plugins/DeflectPlugin.cpp
```diff
@@ -32,6 +32,8 @@
 
 #include <deflect/SizeHints.h>
 
+#include <plugins/extensions/plugins/RocketsPlugin.h>
```
rdumusc: Why included, I don't see it being used in this file..?

plugins/extensions/plugins/RocketsPlugin.h
```diff
@@ -116,12 +118,35 @@ class RocketsPlugin : public ExtensionPlugin
...
     std::unique_ptr<rockets::Server> _rocketsServer;
     using JsonRpcServer = rockets::jsonrpc::Server<rockets::Server>;
-    std::unique_ptr<JsonRpcServer> _jsonrpcServer;
 
     ImageGenerator _imageGenerator;
 
     Timer _timer;
+
+public:
+    std::unique_ptr<JsonRpcServer> _jsonrpcServer;
```
rdumusc: style:
- please keep a single public / protect / private section, in this order
- _ prefix is used by convention for private variables and methods (only)

plugins/extensions/plugins/RocketsPlugin.cpp
```diff
@@ -89,11 +89,13 @@ std::string hyphenatedToCamelCase(const std::string& hyphenated)
 
 namespace brayns
 {
+RocketsPlugin* RocketsPlugin::_instance{nullptr};
```
rdumusc: global variables are evil! ;-) Please find a cleaner way to pass the RocketsPlugin to whoever needs it.

plugins/extensions/plugins/RocketsPlugin.h
```diff
@@ -116,12 +118,35 @@ class RocketsPlugin : public ExtensionPlugin
...
+inline void registerAction(const std::string& name,
+                           std::function<void(Params)> action)
+{
+    RocketsPlugin::instance()->_jsonrpcServer->connect<Params>(name, action);
+}
+
+inline void registerAction(const std::string& name,
+                           std::function<void()> action)
+{
+    RocketsPlugin::instance()->_jsonrpcServer->connect(name, action);
```
rdumusc: if that's where the instance is needed, then the free functions should be made class functions

PR #337, opened 2018-03-05
Fix docker build; add option for OpenDeck module, default OFF

tribal-tec:


favreau:
I was about to do it ;-) Thanks.

PR #338, opened 2018-03-05
Fixed runtime loading of volumes

favreau:
- Fixed runtime loading of volumes
- Moved sample-per-ray to rendering settings

### DIFF
--- a/brayns/common/volume/VolumeHandler.h
+++ b/brayns/common/volume/VolumeHandler.h
@@ -149,8 +149,6 @@ public:
                          const Vector3f& offset);
         ~VolumeDescriptor();
 
-        void clear();
-
         /**
          * @brief Maps the volume to the corresponding _filename
          */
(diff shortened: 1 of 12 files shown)


brayns/common/volume/VolumeHandler.h
```diff
@@ -147,6 +149,8 @@ class VolumeHandler
                          const Vector3f& offset);
         ~VolumeDescriptor();
 
+        void clear();
```
tribal-tec: where's the definition of this function?

plugins/engines/ospray/OSPRayScene.cpp
```diff
@@ -612,23 +612,41 @@ void OSPRayScene::commitVolumeData()
...
+
+        // An empty array has to be assigned to the renderers
+        _ospVolumeDataSize = 0;
+        _ospVolumeData =
+            ospNewData(_ospVolumeDataSize, OSP_UCHAR, 0, _getOSPDataFlags());
+        ospCommit(_ospVolumeData);
+        for (const auto& renderer : _renderers)
+        {
+            OSPRayRenderer* osprayRenderer =
+                dynamic_cast<OSPRayRenderer*>(renderer.get());
```
tribal-tec: std::dynamic_pointer_cast

PR #339, opened 2018-03-06
Add ActionInterface for plugins to register custom notifications and requests

tribal-tec:


plugins/extensions/plugins/jsonSerialization.h
```diff
@@ -444,7 +444,8 @@ void init(brayns::AnimationParameters* a, ObjectHandler* h)
 namespace brayns
 {
 template <class T, class F>
-inline bool from_json(T& obj, const std::string& json, F postUpdateFunc = [] {})
+inline bool from_json_(T& obj, const std::string& json,
```
karjonas: Is there a reason for the added underscore? :thinking:

tribal-tec: It caused ambiguities in the test when reusing the from_json function from below. This one is not meant to be reused anywhere else. I probably should move out it of here for clarity.

brayns/Brayns.h
```diff
@@ -152,7 +155,31 @@ class Brayns
...
+                                        _actionInterface.get())};
+        _extensionPluginFactory.add(plugin);
+        return plugin;
+    }
+
 private:
+    EnginePtr _engine;
+    ParametersManager _parametersManager;
+    ExtensionPluginFactory _extensionPluginFactory;
+    std::shared_ptr<ActionInterface> _actionInterface;
```
rdumusc: When there is a pimpl, in principle all the private members (except virtual functions) should go in it

tests/webAPI.cpp
```diff
@@ -116,16 +144,42 @@ RetVal makeRequest(const std::string& method, const Params& params)
...
 
 BOOST_GLOBAL_FIXTURE(ClientServer);
 
 BOOST_AUTO_TEST_CASE(reset_camera)
 {
-    BOOST_CHECK(makeRequest("reset-camera"));
+    const auto target = Brayns().getEngine().getCamera().getTarget();
+    Brayns().getEngine().getCamera().setTarget({1, 2, 3});
+    makeNotification("reset-camera");
+    BOOST_CHECK_EQUAL(Brayns().getEngine().getCamera().getTarget(), target);
```
rdumusc: Brayns() is not a good method name because it shadows the class constructor. Also, it would be more readable to not do 3 times: Brayns().getEngine().getCamera()

brayns/Brayns.h
tribal-tec: Agreed in principle. This is hopefully just needed for now until we have runtime plugin loading, so this can be private again.

tests/webAPI.cpp
tribal-tec: Will clean this up

PR #349, opened 2018-03-12
Add support for dynamic plugin loading

tribal-tec:


brayns/CMakeLists.txt
```diff
@@ -7,13 +7,15 @@
 add_subdirectory(parameters)
 add_subdirectory(common)
 add_subdirectory(io)
+add_subdirectory(plugin)
```
rdumusc: pluginapi to match lib name?

tests/braynsTestData.cpp
```diff
@@ -126,15 +126,9 @@ BOOST_AUTO_TEST_CASE(render_protein_in_stereo_and_compare)
...
-                          "on",
-                          "--accumulation",
-                          "off",
-                          "--pdb-file",
-                          pdbFile.c_str(),
-                          "--camera",
-                          "stereo"};
+    const char* argv[] = {
+        app,          "--synchronous-mode", "on",       "--accumulation", "off",
+        "--pdb-file", pdbFile.c_str(),      "--camera", "stereo"};
```
rdumusc: cleanup whitespaces

brayns/Brayns.cpp
```diff
@@ -113,6 +108,54 @@ struct Brayns::Impl
...
+#endif
+#ifdef BRAYNS_USE_DEFLECT
+        _extensionPluginFactory.add(
+            std::make_shared<DeflectPlugin>(_engine, this));
+#endif
+    }
+
+    void loadPlugins()
+    {
+#ifdef BRAYNS_USE_OSPRAY
```
karjonas: Is it a problem we need to have ospray enabled to load our plugins?

brayns/CMakeLists.txt
tribal-tec: sure, I renamed it later and noticed we already have pluginS for the built-in ones

tests/braynsTestData.cpp
tribal-tec: damn clangformat. It's shouldn't be my job :(

brayns/Brayns.cpp
tribal-tec: Today nothing works without having ospray anyways. But for your plugins you don't need/see it. We can also use lunchbox::DSO for the library loading. Or some boost stuff.

tests/braynsTestData.cpp
tribal-tec: Can't do it. clangformat likes it this way

tests/ClientServer.h
```diff
@@ -0,0 +1,160 @@
...
+        return responseFuture.get();
+    }
+
+    template <typename Params>
+    void makeNotification(const std::string& method, const Params& params)
+    {
+        client.notify<Params>(method, params);
+
+        wsClient.process(5);
+        for (size_t i = 0; i < 10; ++i)
```
favreau: Magic numbers 5 and 10? Can you define constants for that with an explicit name?

tribal-tec: no :)

PR #375, opened 2018-04-30
Add loader API and registry

tribal-tec:
* Allows plugins to load data from client load RPCs
* Cleanups handling of different data types in LoadDataFunctor
* Extract CircuitLoader from MorphologyLoader
* Removes unused SceneLoader
* Adds support for loading folders from upload-path RPC


tribal-tec:
retest this please

brayns/io/NESTLoader.h
```diff
@@ -51,11 +51,22 @@ namespace brayns
...
 
+    void importFromBlob(Blob&&, Scene&, const Matrix4f&, const size_t) final
+    {
+        throw std::runtime_error("Unsupported");
+    }
+
+    void importFromFile(const std::string&, Scene&, const Matrix4f&,
+                        const size_t) final
+    {
+        throw std::runtime_error("Unsupported");
```
rdumusc: this class also can't load anything? Why is it not calling importCircuit() / importSpikeReport()?

brayns/common/loader/LoaderRegistry.cpp
```diff
@@ -0,0 +1,151 @@
...
+        if (!_isSupported(entry, path))
+            continue;
+
+        auto loader = entry.createLoader();
+
+        if (fs::is_directory(path))
+        {
+            fs::directory_iterator begin(path), end;
+            const int numFiles = std::count_if(begin, end, [](const auto& d) {
+                return !fs::is_directory(d.path());
```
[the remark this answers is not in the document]
tribal-tec: Good catch. Pre-filtering will work.

brayns/io/NESTLoader.h
tribal-tec: exactly, which one to call? The unified load API does not work for some of those 'legacy' loaders. Not even sure if we need that. Hence, for now those loaders are not supported by the 'new' way of loading data. The 'legacy' way of loading is not affected by that, though.

rdumusc: But I don't understand what is the blocking factor, is it just lack of time to do it in this PR? If the problem is choosing between 2 methods just make 2 Loader classes, no..?

tribal-tec: I'm not spending time on something we don't use atm. I leave it to @favreau to either adapt this loader or remove it.

favreau: I would just remove it for now. This will have to go to a plug-in in any case. Don't waste time on that loader.

tribal-tec:
@rdumusc the move of the loading to the loader (doh!) is much nicer in the end. thx for pointing it out!

PR #378, opened 2018-05-02
Add positional cmdline argument for loading data

tribal-tec:


brayns/parameters/ParametersManager.cpp
```diff
@@ -127,7 +154,30 @@ VolumeParameters& ParametersManager::getVolumeParameters()
...
+                   po::command_line_style::allow_short)
+            .allow_unregistered()
+            .run();
+    po::store(parsedOptions, vm);
+    po::notify(vm);
+    delete[] argv;
+
     for (AbstractParameters* parameters : _parameterSets)
-        parameters->set(key, value);
+        parameters->parse(vm);
```
rdumusc: I see this code has been moved as-is from AbstractParameters, but those 20 lines are still almost a copy-paste from the function above. Please try to reuse some code inside a common function.

Note: you also forgot to delete the AbstractParameters::set() declaration in the header file.

PR #379, opened 2018-05-02
Added scene management

favreau:
Introduced the notion of models. 
- A scene has a set of models than can handle various assets such as meshes, point clouds, circuits, etc.
- Each models contains the materials that are required by the geometry it handles.
- Models can manage several instances of the geometry it handles, with a different transformation for every instance.
- Models also contain metadata that are exposed to the HTTP interface via the v1/scene end-point.
- Models provide a simple API to add primitives such as spheres, cones, cylinders, and triangle meshes.


brayns/common/scene/Scene.cpp
```diff
@@ -665,138 +231,65 @@ VolumeHandlerPtr Scene::getVolumeHandler() const
 
 bool Scene::empty() const
 {
-    return _spheres.empty() && _cylinders.empty() && _cones.empty() &&
-           _trianglesMeshes.empty() && !getVolumeHandler();
+    bool empty = true;
+    for (const auto& modelDescriptor : _modelDescriptors)
+        empty = empty && modelDescriptor.getModel()->empty();
```
tribal-tec: you can return false directly if a model is not empty

brayns/common/scene/Scene.cpp
```diff
@@ -984,158 +490,317 @@ void Scene::loadFromCacheFile()
...
+void Scene::setMaterialsColorMap(MaterialsColorMap colorMap)
 {
-    _materials.push_back(material);
-    return _materials.size() - 1;
+    for (auto& modelDescriptors : _modelDescriptors)
+        modelDescriptors.getModel()->setMaterialsColorMap(colorMap);
 }
 
-void Scene::setMaterial(const size_t index, const Material& material)
+Boxf Scene::getBounds()
```
tribal-tec: const function

brayns/common/scene/Scene.cpp
```diff
@@ -984,158 +490,317 @@ void Scene::loadFromCacheFile()
...
+    for (const auto& modelDescriptor : _modelDescriptors)
+        if (modelDescriptor.getEnabled())
+        {
+            bounds.merge(modelDescriptor.getModel()->getBounds());
+            ++nbEnabledModels;
+        }
+    if (nbEnabledModels == 0)
+    {
+        // If no model is enabled. return unity bounding box
+        bounds.merge({0, 0, 0});
```
tribal-tec: why? if nothing is visible, there is actually not box :) Is this needed somewhere not to be empty?

brayns/common/scene/Scene.h
```diff
@@ -132,7 +116,19 @@ class Scene : public BaseObject
...
+    BRAYNS_API virtual ModelPtr createModel(
+        const std::string& name, const ModelMetadata& metadata = {}) = 0;
+
+    /**
+        Removes a geometry group from the scene
+        @param index Index of the model
+      */
+    BRAYNS_API virtual void removeModel(const size_t index) = 0;
+
+    /**registry
```
tribal-tec: rm registry

brayns/common/scene/Scene.h
```diff
@@ -247,121 +219,24 @@ class Scene : public BaseObject
...
-    BRAYNS_API void setSpheresDirty(const bool value) { _spheresDirty = value; }
     /**
      * @internal needed to ensure deletion wrt cyclic dependency
      *           scene<->renderer
      */
     virtual void reset();
 
+    /** @return the current size in bytes of the loaded geometry. */
+    size_t getSizeInBytes() const { return _sizeInBytes; }
+    ModelDescriptors& getModelDescriptors() { return _modelDescriptors; }
```
tribal-tec: const?

brayns/common/scene/Scene.h
```diff
@@ -247,121 +219,24 @@ class Scene : public BaseObject
...
+     * @brief initializeMaterials Initializes materials for all models in the
+     * scene
+     * @param colorMap Color map to use for every individual model
+     */
+    void setMaterialsColorMap(
         MaterialsColorMap colorMap = MaterialsColorMap::none);
 
-    /** @return the current size in bytes of the loaded geometry. */
-    size_t getSizeInBytes() const { return _sizeInBytes; }
+    MaterialPtr getBackgroundMaterial() { return _backgroundMaterial; }
```
tribal-tec: const function

brayns/common/scene/Scene.h
```diff
@@ -387,43 +262,31 @@ class Scene : public BaseObject
...
-    bool _trianglesMeshesDirty;
-    Materials _materials;
-    TexturesMap _textures;
+    ModelDescriptors _modelDescriptors;
+    bool _geometryGroupsDirty{true};
     Lights _lights;
 
     // Volume
-    mutable VolumeHandlerPtr _volumeHandler;
+    VolumeHandlerPtr _volumeHandler{nullptr};
```
tribal-tec: nullptr not needed

brayns/common/scene/Scene.h
```diff
@@ -387,43 +262,31 @@ class Scene : public BaseObject
...
 
     // Simulation
-    AbstractSimulationHandlerPtr _simulationHandler;
+    AbstractSimulationHandlerPtr _simulationHandler{nullptr};
```
tribal-tec: nullptr not needed

brayns/common/scene/Scene.h
```diff
@@ -387,43 +262,31 @@ class Scene : public BaseObject
...
     TransferFunction _transferFunction;
-    CADiffusionSimulationHandlerPtr _caDiffusionSimulationHandler;
-
-    // Scene
-    Boxf _bounds;
+    CADiffusionSimulationHandlerPtr _caDiffusionSimulationHandler{nullptr};
```
tribal-tec: nullptr not needed

brayns/io/MeshLoader.cpp
```diff
@@ -424,29 +340,21 @@ std::string MeshLoader::getMeshFilenameFromGID(const uint64_t gid)
...
-                                    const Matrix4f &, const size_t)
+bool MeshLoader::importMeshFromFile(const std::string&, Scene&, const Matrix4f&,
+                                    const size_t)
 {
     BRAYNS_ERROR << NO_ASSIMP_MESSAGE << std::endl;
     return false;
 }
 
-bool MeshLoader::exportMeshToFile(const std::string &, Scene &) const
+bool MeshLoader::exportMeshToFile(const std::string&, Scene&) const
```
tribal-tec: remove as well? maybe test the build w/o assimp if it works

brayns/io/XYZBLoader.cpp
```diff
@@ -18,14 +18,14 @@
...
-#include "XYZBLoader.h"
-
 #include <brayns/common/log.h>
+#include <brayns/common/scene/Model.h>
 #include <brayns/common/scene/Scene.h>
 #include <brayns/common/utils/Utils.h>
-
 #include <fstream>
 
+#include "XYZBLoader.h"
```
tribal-tec: leave this at the top of the file?

brayns/io/XYZBLoader.cpp
```diff
@@ -39,12 +39,10 @@ std::set<std::string> XYZBLoader::getSupportedDataTypes()
 }
 
 void XYZBLoader::importFromBlob(Blob&& blob, Scene& scene,
+                                const size_t index BRAYNS_UNUSED,
                                 const Matrix4f& transformation,
-                                size_t materialID)
+                                const size_t defaultMaterialId BRAYNS_UNUSED)
 {
-    if (materialID == NO_MATERIAL)
```
tribal-tec: you sure? if it is NO_MATERIAL, it's infinity, and you see no geometry :) Probably you changed that behavior.

brayns/io/algorithms/MetaballsGenerator.cpp
```diff
@@ -536,7 +535,7 @@ void MetaballsGenerator::_buildTriangles(const Vector4fs& metaballs,
 
         auto& vertices = triangles[defaultMaterialId].vertices;
         auto& normals = triangles[defaultMaterialId].normals;
-        auto& colors = triangles[defaultMaterialId].colors;
+        // auto& colors = triangles[defaultMaterialId].colors;
```
tribal-tec: remove?

brayns/io/algorithms/MetaballsGenerator.cpp
```diff
@@ -564,11 +563,13 @@ void MetaballsGenerator::_buildTriangles(const Vector4fs& metaballs,
                 const auto normal = normalize(_edgeVertices[index].normal);
                 normals.push_back(normal);
 
+                /* TODO
```
tribal-tec: same?

brayns/tasks/LoadDataFunctor.cpp
```diff
@@ -74,13 +74,7 @@ void LoadDataFunctor::_performLoad(const std::function<void()>& loadData)
         while (!_lock.try_lock_for(std::chrono::seconds(1)))
             _updateProgress("Waiting for scene access ...", 0.f);
 
-        _updateProgress("Unloading ...", 0.f);
-        Scene& scene = _engine->getScene();
-        scene.unload();
-        _loadDefaultScene = true;
-
         _updateProgress("Loading data ...", LOADING_PROGRESS_STEP);
```
tribal-tec: this should be step 0.f then, and you can remove 1 from LOADING_PROGRESS_STEP

plugins/RocketsPlugin/RocketsPlugin.cpp
```diff
@@ -539,7 +536,8 @@ class RocketsPlugin::Impl
                 _engine->getScene().getTransferFunction());
         _handleGET(ENDPOINT_SCENE, _engine->getScene());
         _handlePUT(ENDPOINT_SCENE, _engine->getScene(),
-                   [](Scene& scene) { scene.commitMaterials(Action::update); });
+                   [](Scene& scene) { scene.markModified(); });
```
tribal-tec: shoudn't be needed, as markModified() is called for all incoming messages.

plugins/RocketsPlugin/SnapshotTask.h
```diff
@@ -77,7 +77,6 @@ class SnapshotFunctor : public TaskFunctor
             *_camera = *_params.camera;
         else
             *_camera = engine.getCamera();
-        _camera->setAspectRatio(float(_params.size.x()) / _params.size.y());
```
tribal-tec: you sure? the snapshot has a different ratio. it's handled in the assignment now?

plugins/RocketsPlugin/jsonSerialization.h
```diff
@@ -267,11 +282,30 @@ inline void init(brayns::Material* m, ObjectHandler* h)
...
+    h->add_property("metadata", &g->_metadata);
+    h->set_flags(Flags::DisallowUnknownKey);
+}
+
 inline void init(brayns::Scene* s, ObjectHandler* h)
 {
-    h->add_property("bounds", &s->getWorldBounds(),
-                    Flags::IgnoreRead | Flags::Optional);
-    h->add_property("materials", &s->getMaterials());
+    auto bounds = s->getBounds();
```
tribal-tec: hmm, the bounds would ideally be cached by the scene, so they don't have to computed here. Let's discuss offline maybe.

plugins/engines/ospray/OSPRayMaterial.h
```diff
@@ -0,0 +1,45 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#ifndef OSPRAYMATERIAL_H
+#define OSPRAYMATERIAL_H
+
+#include <brayns/common/material/Material.h>
+#include <ospray_cpp/Data.h>
```
tribal-tec: not used? in any case, don't use they're cpp API, it's terrible :)

plugins/engines/ospray/OSPRayModel.cpp
```diff
@@ -0,0 +1,414 @@
...
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include "OSPRayModel.h"
+#include "OSPRayMaterial.h"
+
+#include <brayns/common/material/Material.h>
+namespace
+{
+const size_t BOUNDINGBOX_MATERIAL_ID = 9999;
```
tribal-tec: no more than 10000-2 materials? so sad :package:

tests/ClientServer.h
```diff
@@ -45,8 +45,9 @@ const size_t SERVER_PROCESS_RETRIES = 10; /*ms*/
 class ForeverLoader : public brayns::Loader
 {
 public:
-    void importFromBlob(brayns::Blob&&, brayns::Scene&, const brayns::Matrix4f&,
-                        const size_t) final
+    void importFromBlob(brayns::Blob&&, brayns::Scene&, const size_t = 0,
```
tribal-tec: those default values are not needed here. they're already in the base class

brayns/common/scene/Model.cpp
```diff
@@ -0,0 +1,284 @@
...
+                << sizeInBytes / 1048576 << " MB), Bounds: " << _bounds
+                << std::endl;
+}
+
+MaterialPtr Model::getMaterial(const size_t materialId)
+{
+    if (_materials.find(materialId) == _materials.end())
+        throw std::runtime_error("Material " + std::to_string(materialId) +
+                                 " is not registered in the model");
+    return _materials[materialId];
```
karjonas: Double lookup of materialId in map.

brayns/common/scene/Model.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+    for (const auto materialId : materialIds)
+        try
+        {
+            getMaterial(materialId);
+        }
+        catch (const std::runtime_error&)
+        {
+            createMaterial(materialId, std::to_string(materialId));
+        }
```
karjonas: I would prefer to just do a _materials.find() instead of trying and catching exceptions.

hernando: The really fishy thing is that getMaterial doesn't return anything. It should have a different name.

brayns/common/scene/Model.h
```diff
@@ -0,0 +1,241 @@
...
+{
+/**
+* @brief The ModelTransformation struct defines the translation, rotation and
+* scaling parameters to be applied to a model
+*/
+struct ModelTransformation : public BaseObject
+{
+    ModelTransformation() = default;
+    ModelTransformation(ModelTransformation&& rhs) = default;
+    ModelTransformation& operator=(ModelTransformation&& rhs) = default;
```
karjonas: All constructors/assignment are default, maybe remove all?

favreau: We cannot because there is another constructor with a different signature below  :(

karjonas: Okay, I missed that :+1:

brayns/common/scene/Model.cpp
favreau: Fixed. Thanks.

brayns/common/scene/Model.h
```diff
@@ -0,0 +1,241 @@
...
+        _updateValue(_translation, value);
+    }
+    Vector3f& getScale() { return _scale; }
+    void setScale(const Vector3f& value) { _updateValue(_scale, value); }
+    Vector3f& getRotation() { return _rotation; }
+    void setRotation(const Vector3f& value) { _updateValue(_rotation, value); }
+private:
+    Vector3f _translation{0.f, 0.f, 0.f};
+    Vector3f _scale{1.f, 1.f, 1.f};
+    Vector3f _rotation{0.f, 0.f, 0.f};
```
hernando: I guess my opinion doesn't count very much here, but this looks like Euler angles, which are quite cumbersome to specify formally (is this xzx, xyz, xyx, intrinsic, extrinsic, ...?). The interpretation is not even documented here.
Why don't you use a quaternion or an angle, axis representation, which are unambiguous and self-explanatory?

plugins/engines/ospray/OSPRayModel.cpp
karjonas: I regularly use more than that actually. Maybe set it to size_t (-1)?

plugins/RocketsPlugin/jsonSerialization.h
```diff
@@ -267,11 +282,30 @@ inline void init(brayns::Material* m, ObjectHandler* h)
...
+    h->add_property("bounds", &bounds, Flags::IgnoreRead | Flags::Optional);
```
karjonas: Are you sending a pointer to a temporary variable 'bounds' here?

favreau: Yes! :) Fixed ;)

brayns/io/ProteinLoader.cpp
```diff
@@ -434,27 +434,36 @@ bool ProteinLoader::importPDBFile(const std::string& filename,
...
+        size_t i = 0;
+        for (const auto& spheresPerMaterial : spheres)
+        {
+            const auto materialId = spheresPerMaterial.first;
+            auto material =
+                model->createMaterial(materialId, colorMap[i].symbol);
+            material->setDiffuseColor({colorMap[materialId].R / 255.f,
+                                       colorMap[materialId].G / 255.f,
+                                       colorMap[materialId].B / 255.f});
+            for (const auto& sphere : spheresPerMaterial.second)
```
karjonas: Why do you use colorMap[i] and colorMap[materialId] for the same material?

favreau: Fixed!

brayns/common/scene/Model.cpp
favreau: Fixed!

brayns/common/scene/Model.h
karjonas: I agree, I guess we will match these with what the user uses in the UI.

karjonas:
I cannot comment too much on the design since I am not familiar enough with all the pieces but to me it looks fine :)

brayns/common/scene/Model.h
favreau: It's a discussion that comes and go on a regular basis ;) I don't have a strong opinion there, I simply find the translation/scaling/rotation more human readable, and since this is what is exposed to the HTTP interface, I decided to use that representation. I guess the JS guy's opinion also is to consider. I'll ask input from @rolandjitsu

plugins/engines/ospray/OSPRayModel.cpp
favreau: I like that! Done.

brayns/common/scene/Model.h
hernando: This interface doesn't have to match what users see necessarily. This class could use a matrix to encode the transformations and the client could use whatever is preferable depending on the use case, converting it to a matrix at the last step. That would be my choice, in fact.

hernando: I know this discussion is recurrent and I don't what to start it again here. My criticism is that the way the rotation is expressed in this API is unclear (If you give this to two different people, the chances of ending up with a different transformation are quite high). At the very least, it should be documented better. And since it's not so easy to document easily, I suggest to use something which is unambiguous instead.

hernando: Any method for manipulating objects from the GUI in a natural way and relative to the camera position will have to do some transformations, no matter what this API is, and for that you need an algebraic representation (matrix or quaternion), which means that in the end you need to transform the user input to a matrix or quaternion.
If this API describes the rotation as something else, it means that you need to transform the algebraic representation to whatever this API requires.  Any representation can be easily converted to a quaternion or a matrix, but I find the opposite more annoying.
For me these are good reasons to choose a quaternion.

favreau:
Retest this please

PR #380, opened 2018-05-03
Fix #289; fix textures for scivis renderer

### LINKED ISSUE #289: Reduce frequency of progress updates
If a lot of morphologies are loaded, it generates a huge load of messages. Maybe in general all message sends should be throttled. Reuse the image stream fps?

tribal-tec:


plugins/engines/ospray/OSPRayScene.cpp
```diff
@@ -43,11 +43,12 @@ struct TextureTypeMaterialAttribute
...
 };
 
-static TextureTypeMaterialAttribute textureTypeMaterialAttribute[7] = {
+static TextureTypeMaterialAttribute textureTypeMaterialAttribute[8] = {
     {TT_DIFFUSE, "map_kd"},
     {TT_NORMALS, "map_bump"},
+    {TT_BUMP, "map_bump"},
     {TT_SPECULAR, "map_ks"},
-    {TT_EMISSIVE, "map_a"},
+    {TT_EMISSIVE, "map_Ns"},
```
favreau: Any reason why this one has capital N, when all other attributes are lowercase? convention? gltf?

tribal-tec: no, ospray. but they also accept lower-case 'n'. I'll change it

PR #381, opened 2018-05-07
Add support to load data from archives

tribal-tec:


brayns/common/utils/Utils.cpp
```diff
@@ -54,4 +59,134 @@ strings parseFolder(const std::string& folder, const strings& filters)
...
+    return false;
+#endif
+}
+
+bool isSupportedArchiveType(const std::string& extension BRAYNS_UNUSED)
+{
+#ifdef BRAYNS_USE_LIBARCHIVE
+    // No way to get all supported types from libarchive...
+    // Extend this list if you feel your favorite archive type should be here
+    const std::set<std::string> extensions{"zip", "tar.gz", "tgz", "tar.bz2"};
```
rolandjitsu: I guess `.rar` is also a popular archive ext.

brayns/tasks/LoadDataFunctor.cpp
```diff
@@ -43,6 +47,25 @@ LoadDataFunctor::LoadDataFunctor(EnginePtr engine)
...
+            Scope(Blob&& blob, LoadDataFunctor& parent)
+            {
+                fs::create_directories(path);
+
+                extractBlob(std::move(blob), path.string());
+                parent._performLoad([&] { parent._loadData(path.string()); });
+            }
+
+            ~Scope() { fs::remove_all(path); }
+            fs::path path = "/tmp" / fs::unique_path();
```
rdumusc: "/tmp" -> fs::temp_directory_path()? Even better, wouldn't it possible to process the archive in-memory, without writing and re-reading files?

brayns/tasks/UploadBinaryTask.cpp
```diff
@@ -121,7 +122,8 @@ void UploadBinaryTask::_checkValidity(EnginePtr engine)
         if (param.type.empty() || param.size == 0)
             throw MISSING_PARAMS;
 
-        if (!registry.isSupported(param.type))
+        if (!registry.isSupported(param.type) &&
+            !isSupportedArchiveType(param.type))
```
rdumusc: Hmm instead of special handling here (which applies to uploads only?) couldn't this be handled by the registry? What about opening zip files from a filesystem path (not uploaded)?

rdumusc:
Great idea otherwise! :-)

brayns/tasks/LoadDataFunctor.cpp
tribal-tec: Thx for the first one, I was looking for that the whole time!
The latter is only possible if the archive only contains a single, because that's what the loaders support. Otherwise, a virtual filesystem is needed, where there is nothing trivial available.

brayns/tasks/UploadBinaryTask.cpp
tribal-tec: I wanted do that initially, but discarded it at some point. For the remote location, the user can unpack to a folder himself :)
But I'll look at it again, it should be possible, unless I really find the blocker why I didn't do it in the end.

brayns/tasks/LoadDataFunctor.cpp
rdumusc: OK I see... I just remembered that libwebsockets has the ability to server contents directly from a zip file, but that's indeed using a virtual file system, which I don't know much about... probably too complicated for now..!

tribal-tec:
retest this please

tribal-tec:
retest this please

PR #388, opened 2018-05-18
Upgrade to Deflect 1.0,  fix build without Brion

rdumusc:


rdumusc:
retest this please

rdumusc:
Do you have an idea about the unit test that failed twice already in cscs-debug: addModel (SEGFAULT)
? I can't make sense of the log output...

tribal-tec:
Not really... I saw it once too, but the next retrigger worked then. No idea where this comes from, and can be of course not reproduced locally...

rdumusc:
OK... I'll try to retest this please a 3rd time then...

PR #391, opened 2018-05-18
SDF geometries support

karjonas:
A new geometry type called ExtendedSDFGeometries is added. This PR also includes a thickness smoothing algorithm used when loading morphologies. It can be activated by adding '--dampen-branch-thickness-changerate true' when running Brayns. To create SDF geometries when loading a morphology you add '--use-sdf-geometries true'


brayns/io/MorphologyLoader.cpp
```diff
@@ -313,18 +328,290 @@ class MorphologyLoader::Impl
...
+        }
+    }
+
+    /**
+     * @brief _calculateMorphologyTreeStructure Calculates the structure of the
+     * morphology tree by finding overlapping beginnings and endings of the
+     * sections.
+     */
+    MorphologyTreeStructure _calculateMorphologyTreeStructure(
+        const brain::neuron::Sections& sections) const
```
rdumusc: this function is very long and could be refactored into smaller parts

brayns/io/MorphologyLoader.cpp
```diff
@@ -313,18 +328,290 @@ class MorphologyLoader::Impl
...
+        }
+
+        MorphologyTreeStructure mts;
+        mts.sectionTraverseOrder = std::move(sectionOrder);
+        mts.bifurcationParent = std::move(bifurcationParent);
+        return mts;
+    }
+
+    /**
+       * @brief _importMorphologyFromURI imports a morphology from the specified
```
rdumusc: rm @brief _importMorphologyFromURI

brayns/io/MorphologyLoader.cpp
```diff
@@ -313,18 +328,290 @@ class MorphologyLoader::Impl
...
+       * URI
+       * @param uri URI of the morphology
+       * @param index Index of the current morphology
+       * @param transformation Transformation to apply to the morphology
+       * @param material Material that is forced in case geometry parameters
+       * do not apply
+       * @param compartmentReport Compartment report to map to the morphology
+       * @param scene Scene to which the morphology should be loaded into
```
rdumusc: description does not match all params

brayns/io/MorphologyLoader.cpp
```diff
@@ -379,25 +674,37 @@ class MorphologyLoader::Impl
...
+                            const auto& samples = child.getSamples();
+                            const Vector3f sample{samples[0].x(),
+                                                  samples[0].y(),
+                                                  samples[0].z()};
+                            model.addCone(materialId,
+                                          {somaPosition, sample, radius,
+                                           _getCorrectedRadius(samples[0].w() *
+                                                               0.5f),
+                                           0.f, textureCoordinates});
+                        }
```
rdumusc: move to separate function?

brayns/io/MorphologyLoader.cpp
```diff
@@ -461,17 +778,29 @@ class MorphologyLoader::Impl
...
+                    assert(previousRadius >= 0.0f);
                 }
 
                 bool done = false;
-                for (size_t i = step; !done && i < samples.size() + step;
-                     i += step)
+                for (size_t i = step; !done && i < numSamples + step; i += step)
                 {
-                    if (i >= samples.size())
+                    if (i >= numSamples || i == (numSamples - 1))
```
rdumusc: why not just i >= (numSamples - 1) ?

brayns/io/MorphologyLoader.cpp
```diff
@@ -521,31 +848,108 @@ class MorphologyLoader::Impl
...
                                                   {position, target, radius,
-                                                   distance,
+                                                   previousRadius, distance,
                                                    textureCoordinates});
-                            else
-                                model.addCone(materialId,
-                                              {position, target, radius,
-                                               previousRadius, distance,
-                                               textureCoordinates});
+                            }
```
rdumusc: this function is also too long and should be split up

brayns/parameters/GeometryParameters.cpp
```diff
@@ -66,6 +66,9 @@ const std::string PARAM_METABALLS_GRIDSIZE = "metaballs-grid-size";
 const std::string PARAM_METABALLS_THRESHOLD = "metaballs-threshold";
 const std::string PARAM_METABALLS_SAMPLES_FROM_SOMA =
     "metaballs-samples-from-soma";
+const std::string PARAM_DAMPEN_BRANCH_THICKNESS_CHANGERATE =
+    "dampen-branch-thickness-changerate";
```
rdumusc: prefix with "sdf-" ? Otherwise it's hard to know what this applies to.

plugins/engines/ospray/ispc/geometry/ExtendedSDFGeometries.ispc
```diff
@@ -0,0 +1,427 @@
...
+
+inline float mix(const float x, const float y, const float a)
+{
+    return x * (1.f - a) + y * a;
+}
+
+// polynomial smooth min (k = 0.1);
+inline float sminPoly(const float a, const float b, const float k)
+{
+    float h = clamp(0.5 + 0.5 * (b - a) / k, 0.0, 1.0);
```
rdumusc: const

plugins/engines/ospray/ispc/geometry/ExtendedSDFGeometries.ispc
```diff
@@ -0,0 +1,427 @@
...
+
+inline float sdSphere(const vec3f &p, const vec3f c, float r)
+{
+    return length(p - c) - r;
+}
+
+inline float sdCapsule(const vec3f &p, const vec3f a, const vec3f b, float r)
+{
+    vec3f pa = p - a, ba = b - a;
+    float h = clamp(dot(pa, ba) / dot(ba, ba), 0.0, 1.0);
```
rdumusc: const const

plugins/engines/ospray/ispc/geometry/ExtendedSDFGeometries.ispc
```diff
@@ -0,0 +1,427 @@
...
+
+    vec3f invR = 1.f / r.dir;
+    vec3f tbot = invR * (aabbMin - r.org);
+    vec3f ttop = invR * (aabbMax - r.org);
+    vec3f tmin = min(ttop, tbot);
+    vec3f tmax = max(ttop, tbot);
+    vec2f t = make_vec2f(max(tmin.x, tmin.y), max(tmin.x, tmin.z));
+    t0 = max(t.x, t.y);
+    t = make_vec2f(min(tmax.x, tmax.y), min(tmax.x, tmax.z));
+    t1 = min(t.x, t.y);
```
rdumusc: n * const

brayns/common/scene/Model.h
```diff
@@ -268,6 +290,12 @@ class Model
     Boxf _bounds;
     bool _useSimulationModel{false};
 
+    std::map<size_t, std::vector<uint32_t>> _SDFGeometryIndices;
+    std::vector<size_t> _SDFNeighboursFlat;
+    std::vector<std::vector<size_t>> _SDFNeighbours;
+    std::vector<SDFGeometry> _SDFGeometries;
+    bool _SDFGeometriesDirty{false};
```
rdumusc: Not sure, but that's a lot of additions to the Model class, could maybe go into a separate SDF class instanced here..?

brayns/common/scene/Model.h
```diff
@@ -189,6 +190,27 @@ class Model
...
+
+    /** Set the list of neighbours for a SDF geometry
+      @param geometryIdx Index of the geometry
+      @param neighbourIndices Indices of the neighbours
+      */
+    void setSDFGeometryNeighbours(size_t geometryIdx,
+                                  const std::vector<size_t>& neighbourIndices);
+
+    /** Builds the flat list of neighbours for the SDF geometries */
+    void buildSDFGeometryNeighboursFlat();
```
rdumusc: Does this have to be public? Who should call it, why and when?

brayns/common/scene/Model.h
```diff
@@ -189,6 +190,27 @@ class Model
...
+    uint64_t addSDFGeometry(const size_t materialId, const SDFGeometry& geom,
+                            const std::vector<size_t>& neighbours);
+
+
+    /** Set the list of neighbours for a SDF geometry
+      @param geometryIdx Index of the geometry
+      @param neighbourIndices Indices of the neighbours
+      */
+    void setSDFGeometryNeighbours(size_t geometryIdx,
+                                  const std::vector<size_t>& neighbourIndices);
```
rdumusc: How about renaming this function to updateSDFGeometryNeighbours? They are already set in addSDFGeometry() if I'm correct.

Also, this may be a stupid question, but where do I find the geometryIdx to pass to this function?

brayns/common/scene/Model.h
```diff
@@ -189,6 +190,27 @@ class Model
      * now on.
      */
     BRAYNS_API void addModel(ModelPtr model, const Transformations& transform);
+    
+    /**
+      Adds a SDFGeometry to the scene
+      @param materialId Material of the geometry
+      @param geom Geometry to add
+      @param neighbours List of global indices of the geometry neighbours
```
rdumusc: For people like me who are not familiar with SDF terminology, it would great if you could document what neighbours are used for and how I should compute them.

brayns/common/scene/Model.h
tribal-tec: I was wondering in general if this should be in the core at all or rather be a 'real' plugin? Maybe not now, but latest when the circuit loader and friends are moved to a plugin, this shall move as well.

brayns/io/MorphologyLoader.cpp
karjonas: I left this as it is but added some comments since I think the method is actually quite straightforward in its current state. If you disagree I can fix it.

brayns/common/scene/Model.cpp
```diff
@@ -105,10 +105,59 @@ void Model::addModel(ModelPtr model, const Transformations& transform)
...
+
+    switch (geom.type)
+    {
+    case brayns::SDFType::Sphere:
+    {
+        _bounds.merge(geom.center - Vector3f(geom.radius));
+        _bounds.merge(geom.center + Vector3f(geom.radius));
+        break;
+    }
+    case brayns::SDFType::Pill:
```
tribal-tec: use default: instead?

karjonas:
All issues should be fixed now

favreau:
Ultra super cool PR! :)

brayns/common/scene/Model.cpp
karjonas: I added a throw on the default case so that you do not forget to set the bounds if you add a new geometry type.

karjonas:
I also added the textures and timestamp variables to the SDF geometries now so we can get the proper simulation colors.

brayns/common/scene/Model.cpp
```diff
@@ -118,10 +118,61 @@ void Model::addModel(ModelPtr model, const Transformations& transform)
...
+        throw std::runtime_error("No bounds found for SDF type.");
+    }
+
+    return geomIdx;
+}
+
+void Model::updateSDFGeometryNeighbours(
+    size_t geometryIdx, const std::vector<size_t>& neighbourIndices)
+{
+    _sdfGeometryData.neighbours[geometryIdx] = neighbourIndices;
```
rdumusc: doesn't this make the geometry dirty?

brayns/common/scene/Model.cpp
```diff
@@ -118,10 +118,61 @@ void Model::addModel(ModelPtr model, const Transformations& transform)
...
+            std::max(geom.p0.y() + geom.radius, geom.p1.y() + geom.radius),
+            std::max(geom.p0.z() + geom.radius, geom.p1.z() + geom.radius)};
+
+        _bounds.merge(min);
+        _bounds.merge(max);
+        break;
+    }
+    default:
+        throw std::runtime_error("No bounds found for SDF type.");
+    }
```
rdumusc: I would refactor this a bit, moving the logic to get the min/max to the geometry class and keep here only something like:
_bounds.merge(geom.getMin());
_bounds.merge(geom.getMax());
As an added benefit, it lets you unit test independently the geom.getMin/Max() code :-)

brayns/common/scene/Model.cpp
```diff
@@ -118,10 +118,61 @@ void Model::addModel(ModelPtr model, const Transformations& transform)
...
+{
+    _sdfGeometryData.isDirty = true;
+    const uint32_t geomIdx = _sdfGeometryData.geometries.size();
+    _sdfGeometryData.geometryIndices[materialId].push_back(geomIdx);
+
+    assert(_sdfGeometryData.geometries.size() ==
+           _sdfGeometryData.neighbours.size());
+
+    _sdfGeometryData.neighbours.push_back(neighbourIndices);
+    _sdfGeometryData.geometries.push_back(geom);
```
rdumusc: To limit indirection and make the code more readable, the above could moved to a function of sdfGeometryData and have here:
_sdfGeometryData.add(materialId, geom, neighbourIndices);

brayns/common/scene/Model.h
```diff
@@ -272,6 +292,18 @@ class Model
...
     bool _useSimulationModel{false};
 
+    struct SDFGeometryData
+    {
+        std::vector<SDFGeometry> geometries;
+        std::map<size_t, std::vector<uint32_t>> geometryIndices;
+
+        std::vector<std::vector<size_t>> neighbours;
+        std::vector<size_t> neighboursFlat;
+        bool isDirty{false};
```
rdumusc: the dirty flag for other types of data is kept separately, maybe do the same for consistency..?

brayns/common/scene/Model.cpp
karjonas: Yes, fixed

brayns/common/scene/Model.cpp
karjonas: Yes, I created a function that returns the bounding box.

brayns/common/scene/Model.cpp
karjonas: I renamed the variable to _sdf. Do you think it is readable enough then?

brayns/common/scene/Model.h
karjonas: Yes, fixed.

brayns/common/scene/Model.cpp
rdumusc: hmm sure, although my point was not exactly about the variable name. What I was trying to say is more that since those lines of code operate exclusively on the struct members, it could make sense to put them a member function of the struct which is called it from here... but that's fine as well.

PR #392, opened 2018-05-24
Fix submodel instances due to lack of support in OSPRay; cleanup create- and addModel usage

tribal-tec:


favreau:
For the record: https://github.com/ospray/ospray/issues/256

plugins/engines/ospray/OSPRayScene.cpp
```diff
@@ -110,6 +110,11 @@ void OSPRayScene::commit()
                 BRAYNS_INFO << "Committing " << modelDescriptor->getName()
                             << std::endl;
                 _addInstance(_rootModel, impl.getModel(), transformation);
+
+#ifndef OSPRAY_SUPPORTS_SUBMODEL_INSTANCES
```
favreau: From my understanding, OSPRAY_SUPPORTS_SUBMODEL_INSTANCES will apparently never be true :( Do you still want to keep it?

tribal-tec: After I read the ticket of course not :) I'm about to push that.

favreau:
Super cool PR overall, looking forward to using the new feature :)

PR #393, opened 2018-05-24
Added path tracing renderer

favreau:


brayns/tasks/LoadDataFunctor.cpp
```diff
@@ -49,11 +49,11 @@ void LoadDataFunctor::operator()(Blob&& blob)
     {
         struct Scope
         {
-            Scope(Blob&& blob, LoadDataFunctor& parent)
+            Scope(Blob&& b, LoadDataFunctor& parent)
```
tribal-tec: why this change?

favreau: Clang complains about variable shadowing the other one line 62. Gcc is happy though. I had this issue when compiling on MacOSX

tribal-tec: ah ok

plugins/engines/ospray/ispc/render/PathTracingRenderer.ispc
```diff
@@ -0,0 +1,225 @@
...
+
+        // compute ray direction of uniformly random diffuse ray
+        const vec3f raydir =
+            getRandomVector(sample, dg.Ns, self->abstract.super.epsilon);
+
+        // origin of new ray in path is hitpoint of previous ray in path
+        const vec3f hitpoint = dg.P + self->abstract.super.epsilon * dg.Ns;
+        setRay(localray, hitpoint, raydir);
+        localray.t0 = self->abstract.super.epsilon;
+        localray.t = self->abstract.ambientOcclusionDistance;
```
straaljager: I'm not sure if limiting t (the maximum allowed hit distance along ray) to the ambientOcclusionDistance will work the same as setting it to a very large number. This may or may not result in some weird lighting, but may also be faster.

plugins/engines/ospray/ispc/render/PathTracingRenderer.ispc
```diff
@@ -0,0 +1,225 @@
...
+                        // create and trace a shadowray towards the sun
+                        Ray sunRay;
+                        setRay(sunRay, hitpoint, sunSampleDirection);
+                        if (isOccluded(self->abstract.super.model, sunRay))
+                            // take surface color at this bounce's hitpoint into
+                            // account
+                            Kd = Kd * (1.f - self->abstract.shadows);
+                    }
+                    else
+                        Kd = Kd * (1.f - self->abstract.shadows);
```
straaljager: This doesn't seem right, the if/else statement has the same outcome for both the if and else branch (if the surface faces the light source and is not occluded from it, then the hitpoint should receive additional light).

plugins/engines/ospray/ispc/render/PathTracingRenderer.ispc
```diff
@@ -0,0 +1,225 @@
...
+                const varying Light_SampleRes lightSample =
+                    light->sample(light, dg, s);
+                const vec3f radiance = lightSample.weight;
+                if (reduce_max(radiance) > 0.f)
+                {
+                    const vec3f lightDirection = lightSample.dir;
+                    const vec3f sunSampleDirection =
+                        getConeSample(lightDirection, rng,
+                                      self->abstract.softShadows);
+                    const float sunLight =
```
straaljager: I would rename sunLight to sunAngle to avoid confusion and reflect better what it's doing (that was a bad naming choice on my part, sorry)

plugins/engines/ospray/ispc/render/PathTracingRenderer.ispc
```diff
@@ -0,0 +1,225 @@
...
+                        Kd = Kd * (1.f - self->abstract.shadows);
+                }
+            }
+        }
+        mask = mask * Kd;
+    } // end bounces
+
+    // CHECK THIS!!! WITH OR WITHOUT * MASK
+    // results are very different. Color bleeding is much stronger with mask
+    color = accucolor * mask;
```
straaljager: After some digging into path tracing theory, it would be more correct to not multiply the final color by the mask colour again, as it has already been taken into account. Otherwise this will result into too much colour bleeding.

plugins/engines/ospray/ispc/render/PathTracingRenderer.ispc
favreau: I understand, but the ambient occlusion distance can be set to a very large value anyway. And this way, we can change it at runtime.

plugins/engines/ospray/ispc/render/PathTracingRenderer.ispc
favreau: Hum, you're right, I am not 100% sure either of the current implementation. Let's talk tomorrow.

plugins/engines/ospray/ispc/render/PathTracingRenderer.ispc
favreau: Sure, I usually use cosNL for diffuse and cosLR for specular. In this case, only diffuse shading is processed. And Sun is probably a specific case too since we can have many lights. I'll change it to cosNL then, unless you have an objection.

plugins/engines/ospray/ispc/render/PathTracingRenderer.ispc
favreau: Ok. Fixed! :)

PR #394, opened 2018-05-24
Fix crash when starting deflect stream more than once

tribal-tec:


tribal-tec:
retest this please

tribal-tec:
retest this please

PR #395, opened 2018-05-25
Throw exception if model is empty

karjonas:
Empty models cause issues so throw an exception if they are added to the scene.


brayns/common/scene/Scene.cpp
```diff
@@ -111,6 +111,9 @@ void Scene::clearLights()
 
 size_t Scene::addModel(ModelDescriptorPtr model)
 {
+    if (model->getModel().empty())
```
tribal-tec: The same applies for submodels, so please add this to Model::addModel() as well. I'm about to unify this somewhat duplication of code to add models, but for the time being it would be good to have those checks.

PR #402, opened 2018-06-01
Deflect-related improvements: stream & accum frames cmdline, render throttling

tribal-tec:


brayns/Brayns.cpp
```diff
@@ -389,11 +389,23 @@ struct Brayns::Impl : public PluginAPI
...
-        _renderTimer.stop();
+        const auto& params = _parametersManager.getApplicationParameters();
+        const auto fps = params.getImageStreamFPS();
+        const auto delta = _renderTimer.perSecondSmoothed() - fps;
+        if (delta > 0)
+        {
+            const int64_t targetTime = (1. / fps) * 1000.f;
+            std::this_thread::sleep_for(std::chrono::milliseconds(
+                targetTime - _renderTimer.milliseconds()));
+        }
```
rdumusc: This should either be called only when streaming is enabled, or the parameter should be renamed from imageStreamFPS to something more generic (with a way to adjust/disable it when benchmarking performance).

tribal-tec: I don't want to rename it now to not break the web API. It can already be changed on the commandline.

rdumusc: OK, I would still rename it everywhere else, but up to you.

tribal-tec: I added a different parameter for this now. The tests were using the existing FPS to 'disable' streaming to speed up the execution of the tests. It also breaks nothing else at the moment hopefully.

PR #403, opened 2018-06-01
Workaround to OSPRay epsilon issue

favreau:


plugins/engines/ospray/OSPRayRenderer.cpp
```diff
@@ -86,7 +86,7 @@ void OSPRayRenderer::commit()
     ospSet1i(_renderer, "randomNumber", rand() % 10000);
     ospSet1i(_renderer, "spp", rp.getSamplesPerPixel());
     ospSet1i(_renderer, "electronShading", (mt == ShadingType::electron));
-    ospSet1f(_renderer, "epsilon", rp.getEpsilon());
```
tribal-tec: I would still set both epsilons to make it 'work' for the OSPRay renderers

favreau:
Epsilon should be taken from DifferentialGeometry.

PR #405, opened 2018-06-04
Use perceptual diff for more robust image comparison in tests 

tribal-tec:


tests/brayns.cpp
```diff
@@ -142,11 +147,12 @@ BOOST_AUTO_TEST_CASE(defaults)
     BOOST_CHECK(geomParams.getMemoryMode() == brayns::MemoryMode::shared);
 }
 
+#ifdef BRAYNS_USE_FREEIMAGE
 BOOST_AUTO_TEST_CASE(render_two_frames_and_compare_they_are_same)
```
karjonas: Can you move the pdiff-related tests to braynsTestData.cpp or snapshot.cpp to separate them from the others? Then you would not need to use any ifdefs here.

PR #406, opened 2018-06-04
Stricter arguments parsing

karjonas:
This PR make Brayns exit if invoked with incorrect arguments. It also add the possibility to add arguments to plugins.

### DIFF
--- a/brayns/Brayns.cpp
+++ b/brayns/Brayns.cpp
@@ -1085,7 +1085,7 @@ Brayns::Brayns(int argc, const char** argv)
 {
 }
 
-Brayns::~Brayns() {}
+Brayns::~Brayns() = default;
 
 void Brayns::render(const RenderInput& renderInput, RenderOutput& renderOutput)
 {
--- a/brayns/Brayns.cpp
+++ b/brayns/Brayns.cpp
@@ -165,7 +165,7 @@ struct Brayns::Impl : public PluginAPI
             catch (const std::runtime_error& exc)
             {
                 BRAYNS_ERROR << exc.what() << std::endl;
-                exit(0);
+                exit(EXIT_FAILURE);
             }
         }
 #endif
--- a/brayns/Brayns.cpp
+++ b/brayns/Brayns.cpp
@@ -140,11 +140,10 @@ struct Brayns::Impl : public PluginAPI
                 auto createSym = library.getSymbol("brayns_plugin_create");
                 if (!createSym)
                 {
-                    BRAYNS_ERROR << "Plugin '" << pluginName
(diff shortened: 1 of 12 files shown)


brayns/Brayns.cpp
```diff
@@ -143,12 +144,20 @@ struct Brayns::Impl : public PluginAPI
                                  << "' is not a valid Brayns plugin; missing "
                                     "brayns_plugin_create()"
                                  << std::endl;
-                    return;
+                    exit(0);
```
tribal-tec: EXIT_FAILURE

brayns/Brayns.cpp
```diff
@@ -1076,9 +1086,7 @@ Brayns::Brayns(int argc, const char** argv)
 {
 }
 
-Brayns::~Brayns()
-{
-}
+Brayns::~Brayns() {}
```
tribal-tec: = default in the header

brayns/parameters/ApplicationParameters.h
```diff
@@ -82,6 +88,8 @@ class ApplicationParameters : public AbstractParameters
         _updateValue(_synchronousMode, synchronousMode);
     }
 
+    bool getParallelRendering() { return _parallelRendering; }
```
tribal-tec: const function

brayns/parameters/ParametersManager.cpp
```diff
@@ -79,6 +78,7 @@ void ParametersManager::parse(int argc, const char** argv)
     catch (po::error& e)
     {
         BRAYNS_ERROR << e.what() << std::endl;
+        exit(0);
```
tribal-tec: also EXIT_FAILURE

plugins/engines/ospray/OSPRayEngine.cpp
```diff
@@ -35,32 +35,31 @@
...
+        }
+
+        if (parametersManager.getApplicationParameters().getParallelRendering())
+        {
+            argc++;
+            argv.push_back("--osp:mpi");
         }
 
-        ospInit(&argc, newArgv.data());
+        ospInit(&argc, argv.data());
```
tribal-tec: for consistency, argv[0] should be still the app name imo. I don't know if ospray needs that internally or expects it to be there.

brayns/Brayns.cpp
rdumusc: (...or throw to have a single exit point below)

brayns/Brayns.cpp
karjonas: This needs to be in the cpp file due to the pimpl pattern.

karjonas:
Issues fixed and rebased.

PR #413, opened 2018-06-06
Remove ImageMagick, use FreeImage instead

tribal-tec:


plugins/RocketsPlugin/ImageGenerator.cpp
```diff
@@ -70,28 +41,33 @@ ImageGenerator::ImageBase64 ImageGenerator::createImage(
...
+
+        freeimage::MemoryPtr memory(FreeImage_OpenMemory());
+
+        FreeImage_SaveToMemory(fif, image.get(), memory.get(), flags);
 
-        return {blob.base64()};
+        BYTE* pixels = NULL;
+        DWORD numPixels = 0;
+        FreeImage_AcquireMemory(memory.get(), &pixels, &numPixels);
+        return {base64_encode(pixels, numPixels)};
```
rdumusc: shouldn't this whole block go as a function into ImageUtils?

brayns/common/utils/ImageUtils.h
```diff
@@ -0,0 +1,102 @@
...
+
+    BYTE* line = FreeImage_GetBits(dib);
+    for (unsigned y = 0; y < height; ++y, line += pitch)
+        for (BYTE* pixel = line; pixel < line + lineSize;
+             pixel += bytesperpixel)
+            INPLACESWAP(pixel[0], pixel[2]);
+    return true;
+}
+
+ImagePtr getImageFromFrameBuffer(FrameBuffer& frameBuffer)
```
rdumusc: getImageFromFrameBuffer and SwapRedBlue32 should go into ImageUtils.cpp. You can get compilation errors about double definitions if you try using them in two places, unless the function are inlined, but I don't see a good reason to inline them.

PR #414, opened 2018-06-06
Give precedence to later added loaders

karjonas:
This is to make sure that loaders added by plugins are matched before built-in ones.


favreau:
Retest this please

PR #416, opened 2018-06-07
Rename shadows -> shadow_intensity

tribal-tec:


favreau:
Cool. But you should also rename softShadows to shadowSoftness?

tribal-tec:
After talking to @nantille, softShadows is better than shadowSoftness.

favreau:
Ok, thanks for the Info

PR #424, opened 2018-06-11
RPC for requesting objects, schema & version

tribal-tec:
- v1/ prefix for endpoints is gone
- set-<object> replaces <object> notification to update objects
- added get-<object> request to retrieve current object state

### DIFF
--- a/plugins/RocketsPlugin/RocketsPlugin.cpp
+++ b/plugins/RocketsPlugin/RocketsPlugin.cpp
@@ -47,35 +47,46 @@
 
 namespace
 {
-const std::string ENDPOINT_API_VERSION = "v1/";
+// REST PUT & GET, JSONRPC set-* notification, JSONRPC get-* request
 const std::string ENDPOINT_ANIMATION_PARAMS = "animation-parameters";
 const std::string ENDPOINT_APP_PARAMS = "application-parameters";
 const std::string ENDPOINT_CAMERA = "camera";
-const std::string ENDPOINT_FRAME_BUFFERS = "frame-buffers";
 const std::string ENDPOINT_GEOMETRY_PARAMS = "geometry-parameters";
-const std::string ENDPOINT_IMAGE_JPEG = "image-jpeg";
 const std::string ENDPOINT_TRANSFER_FUNCTION = "transfer-function";
 const std::string ENDPOINT_RENDERING_PARAMS = "rendering-parameters";
 const std::string ENDPOINT_SCENE = "scene";
 const std::string ENDPOINT_SCENE_PARAMS = "scene-parameters";
-const std::string ENDPOINT_SIMULATION_HISTOGRAM = "simulation-histogram";
-const std::string ENDPOINT_STATISTICS = "statistics";
 const std::string ENDPOINT_STREAM = "stream";
-const std::string ENDPOINT_STREAM_TO = "stream-to";
-const std::string ENDPOINT_VERSION = "version";
-const std::string ENDPOINT_VOLUME_HISTOGRAM = "volume-histogram";
 const std::string ENDPOINT_VOLUME_PARAMS = "volume-parameters";
 
+// REST GET, JSONRPC get-* request
+const std::string ENDPOINT_STATISTICS = "statistics";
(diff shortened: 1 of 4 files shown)


tribal-tec:
I'm waiting for @rolandjitsu to test this, plus I should bump the version now after breaking the API...

plugins/RocketsPlugin/RocketsPlugin.cpp
```diff
@@ -47,7 +47,6 @@
 
 namespace
 {
-const std::string ENDPOINT_API_VERSION = "v1/";
 const std::string ENDPOINT_ANIMATION_PARAMS = "animation-parameters";
```
rolandjitsu: Let's group these methods/endpoints according to which can be a request or a notification. Maybe also according to which have `set-*`/`get-*`.

tribal-tec: Done

plugins/RocketsPlugin/RocketsPlugin.cpp
```diff
@@ -47,36 +47,47 @@
...
-const std::string METHOD_CHUNK = "chunk";
+const std::string METHOD_SNAPSHOT = "snapshot";
+// METHOD_REQUEST_MODEL_UPLOAD from BinaryRequests.h
+
+// JSONRPC synchronous requests
 const std::string METHOD_GET_INSTANCES = "get-instances";
+const std::string METHOD_IMAGE_JPEG = "image-jpeg";
 const std::string METHOD_INSPECT = "inspect";
-const std::string METHOD_QUIT = "quit";
 const std::string METHOD_REMOVE_MODEL = "remove-model";
```
rolandjitsu: Aren't `'remove-model'` and `'update-model'` jsonrpc notifications?

plugins/RocketsPlugin/RocketsPlugin.cpp
```diff
@@ -47,36 +47,47 @@
...
 const std::string ENDPOINT_SCENE = "scene";
 const std::string ENDPOINT_SCENE_PARAMS = "scene-parameters";
-const std::string ENDPOINT_SIMULATION_HISTOGRAM = "simulation-histogram";
-const std::string ENDPOINT_STATISTICS = "statistics";
 const std::string ENDPOINT_STREAM = "stream";
-const std::string ENDPOINT_STREAM_TO = "stream-to";
+const std::string ENDPOINT_VOLUME_PARAMS = "volume-parameters";
+
+// REST GET, JSONRPC get-* request
+const std::string ENDPOINT_STATISTICS = "statistics";
```
rolandjitsu: So `'statistics'` and `'version'` can now be fetched using a jsonrpc request with `{method: 'get-statistics'}`/`{method: 'get-version'}`?

How about when statistics is broadcasted from the server? Is it broadcasted as `{method: 'get-statistics'}` or `{method: 'statistics'}`?

tribal-tec: Yes, they can be requested as you described.

The broadcast haven't changed, so it's ```statistics``` still

plugins/RocketsPlugin/RocketsPlugin.cpp
tribal-tec: You can use them with notifications if you want. But they can return true/false on success/error.

plugins/RocketsPlugin/RocketsPlugin.cpp
```diff
@@ -739,15 +747,27 @@ class RocketsPlugin::Impl
...
+    void _handleSchemaRPC()
+    {
+        RpcDocumentation doc{"Get the schema of the given endpoint", "endpoint",
+                             "name of the endpoint to get its schema"};
+        _handleRPC<SchemaParam, std::string>(
+            METHOD_SCHEMA, doc, [& schemas = _schemas](const auto& param) {
+                if (schemas.count(param.endpoint) == 0)
+                    throw rockets::jsonrpc::response_error("Endpoint not found",
+                                                           -12347);
+                return schemas[param.endpoint];
```
tribal-tec: json.parse missing here

PR #425, opened 2018-06-11
Fixes related to volume handling

favreau:


favreau:
The handling of volume histogram blocks Brayns for as long as the volume is not fully loaded. This is wrong and Brayns should not currently be the application to provide this kind of data.

brayns/Brayns.cpp
```diff
@@ -225,13 +233,7 @@ struct Brayns::Impl : public PluginAPI
...
         }
 
-        if (_parametersManager.isAnyModified() || camera.isModified() ||
-            scene.isModified())
-        {
-            _engine->getFrameBuffer().clear();
-        }
-
-        _engine->getScene().getTransferFunction().resetModified();
+        scene.getTransferFunction().resetModified();
```
tribal-tec: hmm, this actually already done in commitTransferFunctionData(). we should remove it here then.

brayns/Brayns.cpp
```diff
@@ -191,7 +191,16 @@ struct Brayns::Impl : public PluginAPI
         _extensionPluginFactory.preRender();
 
         auto& scene = _engine->getScene();
+        auto& camera = _engine->getCamera();
+
+        if (_parametersManager.isAnyModified() || camera.isModified() ||
+            scene.isModified() || scene.getTransferFunction().isModified())
```
tribal-tec: Can you test if a markModified() in commitTransferFunctionData() would also work, so that the scene.isModified() is enough here?

plugins/RocketsPlugin/RocketsPlugin.cpp
```diff
@@ -558,7 +558,6 @@ class RocketsPlugin::Impl
 
         _handleFrameBuffer();
         _handleSimulationHistogram();
-        _handleVolumeHistogram();
```
tribal-tec: shouldn't we remove the entire code then? I'm still not sure this is the right thing to do. As long as the client doesn't request the histogram, this shouldn't matter. I will push a change to the VizTools where the states are requested on first access, so this might fix the problem you experienced.

brayns/Brayns.cpp
favreau: OK

brayns/Brayns.cpp
favreau: OK

plugins/RocketsPlugin/RocketsPlugin.cpp
favreau: In a later PR, with volume refactoring.

PR #430, opened 2018-06-14
Clarify beginning of user guide (loading data) 

rdumusc:


brayns/common/utils/ImageUtils.cpp
```diff
@@ -22,6 +22,8 @@
 
 #include <brayns/common/renderer/FrameBuffer.h>
 
+#ifdef BRAYNS_USE_FREEIMAGE
```
rolandjitsu: This is not user guide 😄

rdumusc: That's why this commit is called "Fix build without FreeImage" ;-)

rolandjitsu: I was looking at the PR name 😄

rdumusc: I'm not going to squash the commits, and I don't want to open a separate PR for such a small fix...

PR #438, opened 2018-06-22
Changed viewport to floats in order to increase browser compatibility

favreau:
Some browsers process viewport size as floats rather than uint. Brayns now exposes floats through the ws API but still uses unsigned int internally.

### DIFF
--- a/brayns/parameters/ApplicationParameters.h
+++ b/brayns/parameters/ApplicationParameters.h
@@ -50,10 +50,11 @@ public:
     /** Runtime plugins to load in Brayns::loadPlugins. */
     const std::vector<PluginParam>& getPlugins() const { return _plugins; }
     /** window size */
-    const Vector2ui& getWindowSize() const { return _windowSize; }
+    const Vector2ui getWindowSize() const { return Vector2ui(_windowSize); }
     void setWindowSize(const Vector2ui& size)
     {
-        _updateValue(_windowSize, size);
+        Vector2f value(size);
+        _updateValue(_windowSize, value);
     }
     /** Benchmarking */
     bool isBenchmarking() const { return _benchmarking; }
@@ -89,7 +90,6 @@ public:
     }
 
     bool getParallelRendering() const { return _parallelRendering; }
-
     const std::string& getHttpServerURI() const { return _httpServerURI; }
     void setHttpServerURI(const std::string& httpServerURI)
     {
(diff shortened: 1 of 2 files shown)


rdumusc:
Hmm from my external point of view this change looks wrong. Viewports can't have non-interger sizes and I can't see any reason to do that.

favreau:
They actually can, according to most browsers, hence the change. We did to search on the web and found out that the API used to get sizes of components in web pages are floating point values. Can't remember the API that we looked at, but @rolandjitsu knows them.

rolandjitsu:
@rdumusc sending values from the client side will send doubles, depending on the OS, browser, etc. 

See [getBoundingClientRect()](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) and [DOMRect](https://developer.mozilla.org/en-US/docs/Web/API/DOMRectReadOnly/width), which is what I use to get the current width/height of the user's viewport.

So it might make sense to accept such values.

rdumusc:
No it doesn't, unless I'm missing a part of the picture here. No matter what the JS API is giving you Brayns will never use floating point value. So I can't see what is the benefit of exposing meaningless values...

hernando:
I don't know what the rational behind the DOMRect properties, but they look inspired in a vector graphics world to me. The undebatable truth is that the framebuffer is pixel based. I don't fully understand the diff to be honest, but if your intention is to announce the viewport size as float values because that's what browsers like it's OK. But allowing the client to set a non integer viewport size in Brayns looks definetely wrong. The client should deal with that because it's who knows how to map an discrete size image into whatever the browser uses.

rdumusc:
annoncing the viewport size as float is also wrong ;-)

favreau:
Do we seriously have to debate on that???

rolandjitsu:
@rdumusc and @hernando while I understand your concerns (or not), accepting a float instead of an int does not have any impact on anything. It only removes the need of code duplication from clients (where each client would need to do the rounding). And most of the clients will be browsers, so why not accept the same type that the browser will be sending?

rdumusc:
No we don't have to debate, I have nothing to gain here since it's not my project. Do whatever you want... ;-)

rolandjitsu:
@rdumusc this also addresses an issue we have currently which I fixed on my end, but cannot deploy without breaking production.

hernando:
I was just wondering whether this can bite you later on because I don't see where's the rounding (is it really rounding, or flooring?) occurring now.

rolandjitsu:
@favreau do you just cast to int or do you round anywhere?

favreau:
The default c++ cast is used. Only the integer value remains.

PR #443, opened 2018-06-26
Fixed volume management

favreau:


brayns/Brayns.cpp
```diff
@@ -47,6 +47,7 @@
 
 #include <brayns/tasks/AddModelTask.h>
 
+#include <brayns/pluginapi/ExtensionPlugin.h>
```
tribal-tec: not needed?

brayns/Brayns.cpp
```diff
@@ -193,6 +194,12 @@ struct Brayns::Impl : public PluginAPI
         auto& scene = _engine->getScene();
         auto& camera = _engine->getCamera();
 
+        if (_parametersManager.isAnyModified() || camera.isModified() ||
```
tribal-tec: why moved? the transferfunction.isModified() is not needed if you keep the markModified() in OSPRayScene, no?

plugins/RocketsPlugin/RocketsPlugin.cpp
```diff
@@ -733,6 +733,8 @@ class RocketsPlugin::Impl
     {
         auto& params = _parametersManager.getVolumeParameters();
         auto postUpdate = [this](VolumeParameters&) {
+            // Cleanup existing volume data
```
tribal-tec: rm?

plugins/engines/ospray/OSPRayScene.cpp
```diff
@@ -271,68 +270,81 @@ bool OSPRayScene::commitTransferFunctionData()
         ospCommit(impl);
     }
     _transferFunction.resetModified();
-    markModified();
```
tribal-tec: see comment above, to me this can stay as it was

PR #445, opened 2018-06-27
Fixing problems with clipping planes

hernando:


rolandjitsu:
I feel the refactoring was unnecessary, as the API is likely to change (as a consequence of the meeting we had today) for both the way the camera and rendering params is handled.

Wouldn't it have been enough to just fix the constraint of the clipping planes?

hernando:
I know the API is going to change, but you can consider this as an exercise to get familiar with the code base as well.
The refactoring was necessary to allow switching the camera at runtime because in order to create the OSPRay camera object you need to be able to translate a CameraType into a camera type name. That was only doable with the rendering settings object, but in fact the information is global, so I preferred to provide static functions for that rather than making the camera depend on the rendering settings (which is a false dependency) only for that.
Some of the functions could have been static in the first place anyway.

rolandjitsu:
@hernando feel free to close this if you plan on having a different PR.

hernando:
Too many things changed, I will create a new PR

PR #447, opened 2018-06-29
Add support for streamlines

karjonas:


brayns/common/scene/Model.h
```diff
@@ -219,6 +220,22 @@ class Model
...
+      */
+    BRAYNS_API StreamlinesMap& getStreamlines() { return _streamlines; }
+    /**
+      Adds a streamline to the model
+      @param materialId Id of the material for the streamline
+      @param vertices Vertices of the streamline
+      @param colors Colors of the vertices (RGBA)
+      @param radii Radius of the vertices (optional)
+      */
+    BRAYNS_API void addStreamline(const size_t materialId,
```
tribal-tec: using the Streamlines struct with a custom constructor if needed?

favreau:
Please add the test for streamlines in Model::empty(), and add a comment to the getBounds (they are currently only used for volumes, and are is meant to disappear)

PR #448, opened 2018-07-02
Add SharedData- and BrickedVolume support

tribal-tec:
- use OSPRay volume and transfer function to support volume rendering
- add VolumeLoader for *.raw and *.mhd volumes


plugins/engines/ospray/OSPRayScene.h
```diff
@@ -50,32 +50,33 @@ class OSPRayScene : public Scene
     bool commitLights() final;
 
     void commitSimulationData();
-    void commitVolumeData();
+    bool commitVolumeData();
```
rdumusc: this return value is a bit obscure; I would imagine that it indicates a possible failure, but after checking the code that's not the case... I think a separate getter for rebuildScene is preferable

plugins/engines/ospray/ispc/render/utils/AbstractRenderer.ispc
```diff
@@ -347,83 +351,111 @@ inline varying vec4f
...
+            //                    shadowIntensity +=
+            //                        getVolumeShadowContribution(self,
+            //                        lightRay, sample);
+            //                }
+            //                const float giAttenuation =
+            //                    1.f - shadowIntensity * self->shadows;
+            //                voxelColor.x = voxelColor.x * giAttenuation;
+            //                voxelColor.y = voxelColor.y * giAttenuation;
+            //                voxelColor.z = voxelColor.z * giAttenuation;
+            //            }
```
rdumusc: lots of commented code here, is there a good reason not to delete it?

brayns/io/VolumeLoader.cpp
```diff
@@ -0,0 +1,173 @@
...
+        break;
+    case DataType::INT16:
+        dataRange = {std::numeric_limits<int16_t>::min(),
+                     std::numeric_limits<int16_t>::max()};
+        break;
+    case DataType::INT32:
+        dataRange = {std::numeric_limits<int32_t>::min() / 100,
+                     std::numeric_limits<int32_t>::max() / 100};
+        break;
+    }
```
rdumusc: this can easily be a helper function

brayns/common/volume/BrickedVolume.h
```diff
@@ -0,0 +1,40 @@
...
+{
+public:
+    BrickedVolume(const Vector3ui& dimension, const Vector3f& spacing,
+                  const DataType type)
+        : Volume(dimension, spacing, type)
+    {
+    }
+
+    virtual void setBrick(void* data, const Vector3ui& position,
+                          const Vector3ui& size) = 0;
```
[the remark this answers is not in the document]
tribal-tec: no, can be different for each brick if desired

plugins/engines/ospray/OSPRayScene.h
tribal-tec: Doesn't make it better imo. That functions is private anyways, I'll do that first and we can discuss if you like.

plugins/engines/ospray/ispc/render/utils/AbstractRenderer.ispc
tribal-tec: @favreau has to adapt the shaders to use the ospray volumes.

brayns/common/volume/BrickedVolume.h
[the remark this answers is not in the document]
rdumusc: that's interesting, I'd like to learn how Ospray handles that. I wonder what happens in case of overlapping bricks.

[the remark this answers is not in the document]
tribal-tec: There's no overlap, it just overrides potential existing voxels.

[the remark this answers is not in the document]
tribal-tec: Have to const_cast it then to pass it to ospray... Though they use const data internally behind the API...

tribal-tec:
The default camera position has changed somehow, that's why the tests fail. Will look into that.

favreau:
VolumeHandler h and cpp can be removed, right?

tribal-tec:
Yes, I forgot the header file somehow. Thx for the hint!

plugins/engines/ospray/OSPRayScene.cpp
```diff
@@ -78,12 +78,12 @@ OSPRayScene::~OSPRayScene()
 void OSPRayScene::commit()
 {
     const bool rebuildScene = isModified();
+    const bool addRemoveVolumes = _commitVolumeData();
```
rdumusc: Without knowing too much, I still find the logic here convoluted. I don't know why the isModified() is checked before _commitSimulationData and commitTransferFunctionData and what's so different with _commitVolumeData... Couldn't they just set the isModified() flag as needed, and then we would "ideally" have a commit() function that looks like this:
```
void OSPRayScene::commit()
{
    _commitVolumeData();
    _commitSimulationData();
    _commitTransferFunctionData();
    if (isModified())
        _rebuildScene();
}
```

brayns/io/VolumeLoader.cpp
```diff
@@ -0,0 +1,179 @@
...
+        {
+            boost::filesystem::path basePath(filename);
+            path = boost::filesystem::canonical(path, basePath.parent_path());
+        }
+        volumeFile = path.string();
+    }
+    else
+    {
+        dimensions = _volumeParameters.getDimensions();
+        spacing = _volumeParameters.getElementSpacing();
```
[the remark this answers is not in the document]
tribal-tec: the parameters are the fallback if you don't have an mhd. They should be updated at best today if an mhd was provided.

plugins/engines/ospray/OSPRayScene.cpp
tribal-tec: the modified state is also evaluated outside after this function to clear the framebuffer for instance. This happens if for instance the transfer function has changed, hence _commitTransferFunctionData() updates the modified state in the scene. But it does not require a rebuild of the scene, so the state is stored explicitly before.

brayns/common/scene/Model.h
```diff
@@ -324,6 +335,9 @@ class Model
     bool _sdfGeometriesDirty{false};
     bool _instancesDirty{true};
 
+    Volumes _volumes;
+    bool _volumesDirty{true};
```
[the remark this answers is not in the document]
rdumusc: I'm surprised that things are dirty by default but OK. Just FYI there is also bool _sdfGeometriesDirty{false};

PR #452, opened 2018-07-04
Expose SDF attributes to WS API

favreau:


plugins/RocketsPlugin/jsonSerialization.h
```diff
@@ -450,6 +450,11 @@ inline void init(brayns::GeometryParameters* g, ObjectHandler* h)
     h->add_property("memory_mode", &g->_memoryMode, Flags::Optional);
     h->add_property("circuit_configuration", &g->_circuitConfiguration,
                     Flags::Optional);
+    h->add_property("dampen_branch_thickness_changerate",
```
karjonas: Why not **morphology**_dampen_branch_thickness_changerate and **morphology**_use_sdf_geometries?

rolandjitsu: This is a very long param name 😄  But these geometry params have to change as well (possibly remove them altogether), since most of these should be plugin specific and should not be in the core.

favreau: @karjonas, that looked really long with the morphology prefix, and since it's already in the geometry parameters, it sounded a bit like duplication. Ok to put it back :) 
@rolandjitsu, Sure, this will eventually move to plugins, typically for simulation, but I had a recent demand for that functionality, and for morphologies only, not circuits. I think it's still useful to have it in the core for now. Ok for you?

tribal-tec: The SDF geometries are a core feature, but those two parameters are used by the morphology loader only. So maybe, we can reuse the property map stuff here as well for loaders to extend the core properties. Is not enough today to specify those flags at startup on the commandline?

rolandjitsu: @favreau sure. But the sooner we move it to plugins the better (to keep the core as clean as possible) 😉

favreau:
Thanks for your comments, they are showing that adding those attributes at this point adds more confusion than anything else. I'll close the PR. The reason for the 'urgency' was to offer the choice of the visual representation of the morphologies to the scientists, without having to restart Brayns. In practice, I actually think this should even be an attribute of the morphology loader since you might want to have different representations of the same morphology, without having to manipulate 'global' attributes. Anyway, plug-in feature in any case :-)

PR #453, opened 2018-07-05
Explicitly handle image_stream_fps == 0 to disable streaming

tribal-tec:


tribal-tec:
It can be used to disable the streaming. Is useful if the braynsViewer is used where the images congest the link.

PR #459, opened 2018-07-10
Introduce renderer-specific properties

tribal-tec:
* new web API: set-/get-renderer-params which returns the properties of the
  current renderer
* hardcoded renderer types are gone, new renderers are identified by string
* new PropertyObject base class of Renderer keeps properties per type
* only one single renderer exists in Brayns, type switching happens on the
  OSPRay level only
* resolved cyclic dependency scene <-> renderer


tests/perf/renderingSettings.cpp
```diff
@@ -51,8 +52,12 @@ BOOST_AUTO_TEST_CASE(default_scene_benckmark)
     timer.stop();
     reference = timer.milliseconds();
 
+    auto& renderer = brayns.getEngine().getRenderer();
+
     // Shadows
-    params.getRenderingParameters().setShadowIntensity(1.f);
+    auto props = renderer.getPropertyMap();
+    props.updateProperty("shadows", 1.f);
+    renderer.updateProperties(props);
```
rdumusc: I haven't been part of the technical discussions, but this first usage example does not convince me that the change is an improvement:
1) The code is longer
2) Addressing by string is error-prone
3) I'm not sure if type safety is guaranteed; I have the impression that props.updateProperty("shadows", "no_a_number"); would compile just fine, or am I wrong and that's the reason that I can see no test for that?
4) It seem to me that internal properties of the Ospray shaders are being exposed directly throughout the application all the way to the web interface, completely breaking the encapsulation. If a shader in Ospray changes, all user code that relies on that magic string has to be fixed / recompiled / redeployed.

brayns/Brayns.cpp
```diff
@@ -865,45 +864,46 @@ struct Brayns::Impl : public PluginAPI
...
-        renderParams.setAmbientOcclusionStrength(aaStrength);
+        auto& renderer = _engine->getRenderer();
+        if (!renderer.hasProperty("aoWeight"))
+            return;
+
+        auto aoStrength = renderer.getProperty<float>("aoWeight");
+        aoStrength -= 0.1f;
+        if (aoStrength < 0.f)
+            aoStrength = 0.f;
+        renderer.updateProperty("aoWeight", aoStrength);
```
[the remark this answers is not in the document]
tribal-tec: Strictly speaking, those actions should not be in the core, hence it becomes client code where those strings are either hardcoded or better, are taken from the schema. This is the case for the web clients. This change just tries to avoid breaking existing behavior. I can define a string constant here, but it's still wrong :)

tests/perf/renderingSettings.cpp
tribal-tec: That's the entire point: custom shaders have custom attributes, so they can and should no longer be hardcoded in the core. If the shader changes, it should be directly reflected on the client side (btw, the shader will not change really, but there will be new shaders from any plugin/module).

With the schema in place, having the properties for a specific shader being exposed is doable for the clients. Used here w/o any schema checks which tells the valid names and types, this looks wrong, agreed.

I addressed point 3, where the updated value must have the same type that the property was registered with.

rdumusc: OK I see... if the shaders are moved to plugins then then PropertyMap is a way to provide reflexion for arbitrary C++ classes... in that case the string keys make sense. Then yes, make sure the intended usage is well tested:
PropertyMap is filled by plugin and only existing properties can be updated by clients in a type safe-manner.

PR #462, opened 2018-07-12
Introduce camera-specific properties and runtime type switching

tribal-tec:


brayns/common/PropertyMap.h
```diff
@@ -69,14 +69,34 @@ class PropertyMap
...
+        Property(const std::string& name_, const std::string& label_,
+                 const T& value, const std::pair<T, T>& limits)
             : name(name_)
-            , title(title_)
+            , label(label_)
             , type(_getType<T>())
             , _data(value)
-            , _min(min_)
-            , _max(max_)
+            , _min(limits.first)
```
rolandjitsu: Should it not be just `limit` and `limit.min`/`limit.max`? `limits` would implies there is a set of limit pairs.

plugins/RocketsPlugin/jsonPropertyMap.h
```diff
@@ -80,21 +80,42 @@ void _addPropertySchema(const PropertyMap::Property& prop,
...
-        jsonSchema["minimum"] = prop.min<T>();
-    else
-        jsonSchema.AddMember(StringRef("minimum"), prop.min<T>(),
+
+    Document jsonSchema;
+    if (prop.enums.empty())
+    {
+        auto value = prop.get<T>();
+        jsonSchema =
+            staticjson::export_json_schema(&value, &schema.GetAllocator());
```
rolandjitsu: Could make a const out of `schema.GetAllocator()`?

PR #474, opened 2018-07-25
Add python client library

tribal-tec:
* 'building', testing and deployment on PyPIvia travis-ci
* documentation hosting on readthedocs


python/brayns/client.py
```diff
@@ -0,0 +1,226 @@
...
+
+        if in_notebook():
+            self._add_widgets()  # pragma: no cover
+
+    def __str__(self):
+        """Return a pretty-print on the currently connected Brayns instance."""
+        # pylint: disable=E1101
+        version = 'unknown'
+        if self.version:
+            version = '.'.join(str(x) for x in [self.version.major, self.version.minor,
```
rolandjitsu: Semver should be `<major>.<minor>.<patch>`, the revision should probably be separate.

python/brayns/client.py
```diff
@@ -0,0 +1,226 @@
...
+        :param tuple size: (width,height) of the resulting image
+        :param str format: image type as recognized by FreeImage
+        :param int quality: compression quality between 1 (worst) and 100 (best)
+        :param int samples_per_pixel: samples per pixel to increase render quality
+        :return: the PIL image of the current rendering, None on error obtaining the image
+        :rtype: :py:class:`~PIL.Image.Image`
+        """
+        args = locals()
+        del args['self']
+        result = self.snapshot(response_timeout=None, **{k: v for k, v in args.items() if v})
```
rolandjitsu: I imagine `.snapshot()` is one of the dynamically generated fns. If that's the case, could this `.image()` fn accept all that the snapshot would (renderer, etc.)?

python/brayns/client.py
```diff
@@ -0,0 +1,226 @@
...
+        url = 'https://bbp-brayns.epfl.ch?host=' + self.url()
+        webbrowser.open(url)
+
+    def _check_version(self):
+        """Check if the Brayns' version is sufficient enough."""
+        status = utils.http_request(HTTP_METHOD_GET, self.url(), 'version')
+        if status.code != HTTP_STATUS_OK:
+            raise Exception('Cannot obtain version from Brayns')
+
+        minimal_version = '0.7.0'
```
rolandjitsu: Maybe we could keep this min version in the same version file where we have the python version for the lib?

python/brayns/rpcclient.py
```diff
@@ -0,0 +1,232 @@
...
+
+        :param str method: name of the method to invoke
+        :param dict params: params for the method
+        :param int response_timeout: number of seconds to wait for the response
+        :return: result or error of RPC
+        :rtype: dict
+        :raises Exception: if request was not answered within given response_timeout
+        """
+        data = dict()
+        data['jsonrpc'] = "2.0"
```
rolandjitsu: Make a var for the rpc version `JSON_RPC_VERSION = "2.0"`?

python/.gitignore
```diff
@@ -1,6 +1,12 @@
-.idea*
-env/
+.coverage
+.idea/
+brayns.egg-info/
+.installed
+pycodestyle.txt
```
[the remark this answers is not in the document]
tribal-tec: That's the output file telling you about code style violations. The settings are in setup.cfg and in pylintrc.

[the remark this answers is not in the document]
rolandjitsu: Ok, the names were misleading 😄

PR #479, opened 2018-07-28
Added rotation center to model transformation

favreau:


brayns/io/MorphologyLoader.cpp
```diff
@@ -76,6 +76,7 @@ class MorphologyLoader::Impl
      */
     bool importMorphology(const servus::URI& source, Model& model,
```
karjonas: Throw exception on failure and return structure.

favreau: In the end, the only thing I need to return is a soma position, meaning I don't need a structure there.

brayns/common/Transformation.h
```diff
@@ -75,14 +81,16 @@ class Transformation : public BaseObject
...
 
     SERIALIZATION_FRIEND(Transformation)
 };
 inline Transformation operator*(const Transformation& a,
                                 const Transformation& b)
 {
     const auto matrix = a.toMatrix() * b.toMatrix();
-    return {matrix.getTranslation(), a.getScale() * b.getScale(), matrix};
+    return {matrix.getTranslation(), a.getScale() * b.getScale(), matrix,
+            a.getRotationCenter()};
```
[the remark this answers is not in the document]
favreau: The proper way is to remove the transformation from the add_model rpc call. Then the user has to explicitly apply a transformation to a model that has already been loaded. Let me see what I can do there, not sure if this is that simple with regards to the dynamicity of rpc calls.

PR #485, opened 2018-08-02
Update rapidjson to fix gcc 8 compilation in Docker image

tribal-tec:


karjonas:
Just wondering, is there any reason why rapidjson is not a git submodule as the other dependencies usually are?

karjonas:
Makes sense. Probably now we can then remove the "-Wno-unknown-warning-option" I had to add in some places to fix clang compilation.

PR #488, opened 2018-08-02
Exposed volume alpha correction parameters to WS API

favreau:


plugins/engines/ospray/ispc/render/SimulationRenderer.ispc
```diff
@@ -217,43 +336,65 @@ inline vec4f getVolumeContribution(
...
                 dg.epsilon = epsilon;
                 dg.Ns = gradient;
+            }
 
-                gradient =
-                    safe_normalize(volume->computeGradient(volume, point));
+            if (shadingEnabled)
+            {
+                uniform int i = 0;
+                //                for (uniform int i = 0; i <
```
tribal-tec: ?

PR #506, opened 2018-08-10
Add model properties to allow updates on model data from plugins

tribal-tec:


plugins/RocketsPlugin/RocketsPlugin.cpp
```diff
@@ -999,6 +1005,96 @@ class RocketsPlugin::Impl
                       buildJsonRpcSchemaRequest<ModelDescriptor, bool>(desc));
     }
 
+    void _handleGetModelProperties()
+    {
+        const RpcParameterDescription desc{
+            METHOD_GET_MODEL_PROPERTIES,
+            "Get the properties of the given model", "ids",
```
rolandjitsu: I thought that get model properties is for a single model:
```
RPC get-model-properties {id: 'some model id'}
```

plugins/RocketsPlugin/jsonSerialization.h
```diff
@@ -126,6 +137,25 @@ STATICJSON_DECLARE_ENUM(brayns::TextureType,
...
+    // beware that this should only be used for creating JSON RPC schema which
+    // uses PropertyMap as an argument and a oneOf list is not possible. To
+    // properly serialize a property map to/from JSON, use from_json/to_json
+    // from jsonPropertyMap.h directly.
+    h->set_flags(Flags::DisallowUnknownKey);
+}
+inline void init(brayns::ModelProperties* s, ObjectHandler* h)
+{
+    h->add_property("id", &s->modelID);
+    h->add_property("properties", &s->properties);
```
rolandjitsu: Are the params for set model props:
```js
{
  id: 1,
  properties: {
    ping: true
  }
}
```
or:
```js
{
  id: 1,
  ping: true
}
```
?

Second would make more sense, unless that's difficult to implement.

plugins/RocketsPlugin/RocketsPlugin.cpp
tribal-tec: Yes, it is. Have to fix this description.

plugins/RocketsPlugin/jsonSerialization.h
tribal-tec: It's first option for now. Ideally, the RPC would have 2 arguments, but that's not supported right now in Rockets iirc. The second option is quite unnatural, as the 'id' is not a property of the model properties, it's only needed to find the model.

rolandjitsu: Ok. Does this apply for broadcast and get or is it only the set?

tribal-tec: set- and broadcast. get- is just the properties, the ID is the input for the get- request.

plugins/RocketsPlugin/RocketsPlugin.cpp
rolandjitsu: Sounds good.

ppodhajski:
retest this please

PR #507, opened 2018-08-10
Integrate RapidJSON as a submodule

tribal-tec:
Though being a header-only library, copying the files here is a pain when
clang-format touches those files and breaks warning flags from macros.


ppodhajski:
retest this please

PR #510, opened 2018-08-13
Changed warning trace to error if color scheme by layer is choosen for MVD2 circuits

favreau:


brayns/io/CircuitLoader.cpp
```diff
@@ -307,10 +307,10 @@ class CircuitLoader::Impl
         }
         catch (...)
         {
-            BRAYNS_WARN << "Only MVD2 format is currently supported by Brion "
```
tribal-tec: Just print the message if ' Color scheme by layer' was requested? Then it should be an error.

favreau: Fair enough.

tribal-tec: Now the commit message and the PR message don't match anymore.

karjonas: Why do you do this check in the try/catch and not before? This hides other exceptions then.

PR #514, opened 2018-08-14
Use plugin name as first argument for extension plugins.

karjonas:
This makes it conform with how argc/argv usually works. It will also make it easier for plugins that use boost and similar for parsing input arguments.


brayns/Brayns.cpp
```diff
@@ -147,10 +147,12 @@ struct Brayns::Impl : public PluginAPI
                         "brayns_plugin_create()");
                 }
 
+                std::vector<std::string> tmpArgs = pluginParam.arguments;
```
tribal-tec: auto

PR #515, opened 2018-08-15
Add point size property to point cloud models

tribal-tec:


brayns/common/PropertyMap.h
```diff
@@ -146,6 +156,7 @@ class PropertyMap
         const boost::any _min;
         const boost::any _max;
         bool _readOnly{false};
+        std::function<void(const Property&)> _cb;
```
hernando: Another better name? When reading the code in set it's not fully clear what _cb means.

brayns/common/scene/Model.h
```diff
@@ -186,16 +186,14 @@ class Model
...
-    void setBounds(const Boxf& box) { _bounds = box; }
     /**
         Returns spheres handled by the Model
     */
-    BRAYNS_API SpheresMap& getSpheres() { return _spheres; }
+    BRAYNS_API SpheresMap& getSpheres()
+    {
+        _spheresDirty = true;
+        return _spheres;
+    }
```
hernando: This looks really bizarre. You're assuming that the sphere map is going to be changed just be getting it?
At least you should add a const method that doesn't do that and returns a const SphereMap& and document the side effect of calling the non const method. Anyway, if you find a way to move the dirty flag to the SpheresMap, that's even better (I'm suggesting this without knowing the type of SpheresMap)

brayns/common/scene/Model.h
```diff
@@ -261,9 +267,16 @@ class Model
     */
     BRAYNS_API TrianglesMeshMap& getTrianglesMeshes()
```
hernando: getTriangleMeshes sound more English to me (and easier to pronounce)

brayns/common/scene/Model.h
```diff
@@ -261,9 +267,16 @@ class Model
...
     {
+        _trianglesMeshesDirty = true;
         return _trianglesMeshes;
     }
 
+    template <typename T>
+    void updateTrianglesMeshesBounds(const T& value)
+    {
+        _trianglesMeshesBounds.merge(value);
+    }
```
hernando: Why is this special? I don't see any apparent reason for the treating triangle meshes differently.

brayns/io/XYZBLoader.cpp
```diff
@@ -98,25 +98,37 @@ ModelDescriptorPtr XYZBLoader::importFromBlob(
...
+        if (auto modelDesc_ = modelDesc.lock())
+        {
+            const auto newRadius = property.template get<float>();
+            for (auto& sphere : modelDesc_->getModel().getSpheres()[materialId])
+                sphere.radius = newRadius;
+        }
+    });
+    PropertyMap properties;
+    properties.setProperty(radiusProperty);
+    modelDescriptor->setProperties(properties);
```
hernando: How difficult is to support a different radius for each sphere? If the input is 3D vectors you use this code path and if it's 4D you take the forth coordinate as the radius. That would be useful for many use cases.

plugins/engines/ospray/OSPRayScene.cpp
```diff
@@ -84,7 +84,24 @@ void OSPRayScene::commit()
     commitTransferFunctionData();
 
     if (!rebuildScene && !addRemoveVolumes)
-        return;
+    {
+        // copy the list to avoid locking the mutex
+        bool doUpdate = false;
+        auto modelDescriptors = _modelDescriptors;
```
hernando: If there's a race condition, the copy must still be thread-safe.

plugins/engines/ospray/OSPRayScene.cpp
```diff
@@ -138,13 +155,28 @@ void OSPRayScene::commit()
...
+            {
+                // scale and move the unit-sized bounding box geometry to the
+                // model size/scale first, then apply the instance transform
+                const auto& modelBounds =
+                    modelDescriptor->getModel().getBounds();
+                Transformation modelTransform;
+                modelTransform.setTranslation(modelBounds.getCenter() /
+                                                  modelBounds.getSize() -
+                                              Vector3f(0.5f));
+                modelTransform.setScale(modelBounds.getSize());
```
hernando: I don't quite understand what you're doing here.
- The geometry of the unit-sized bb, is centered at 0, 0, 0, or 0.5, 0.5, 0.5?
- modelBound.getSize() is returning a 3d vector or a float?
As I see it, the bounding box gets distorted in funny ways is the model is much larger in one dimension that the other two. Was it already the case? If not I-d prefer another solution that doesn't make bounding boxes look deformed.

brayns/common/scene/Model.h
tribal-tec: I don't want to expose the spheres etc. like this at all like that, but that's the way it is designed by the original author. I just corrected that a non-const access potential modifies the spheres. A const version is not needed atm, but I can add it for sure.

brayns/common/scene/Model.h
tribal-tec: Not my concern here. Feel free to change it.

brayns/common/scene/Model.h
tribal-tec: Again, not changing anything which is not my concern here.

brayns/io/XYZBLoader.cpp
tribal-tec: That's not feasible from the UI, and the radius is not part of the input file,  hence every sphere hast the same radius.

plugins/engines/ospray/OSPRayScene.cpp
tribal-tec: ? It's safe against add/remove of models?!

plugins/engines/ospray/OSPRayScene.cpp
tribal-tec: There is no deformation, be assured.

brayns/common/scene/Model.h
hernando: Well, since you're modifying the code I thought you would know the answer.

brayns/io/XYZBLoader.cpp
hernando: In fact, I'm suggesting to extend the input file, which as far as I know is just a text.
The UI could expose a multiplicative factor, not the absolute radius to be used. That way you could still have explicit radius in the file. As a matter of fact, the multiplicative factor is something useful for animating groups of spheres (e.g. to show a selection when colors are already used for something else)

brayns/common/scene/Model.h
karjonas: I think it would be safe to remove this function and calculate the bounds in the updateBounds function instead. It needs to be tested more though so maybe it should be done in another PR.

PR #519, opened 2018-08-16
Fix broadcast of animation params and scene updates

tribal-tec:


tribal-tec:
Not right now. The different parts (models, clip planes, simulation) of the scene which require a broadcast have to be properly identified and handled.

PR #523, opened 2018-08-16
Do not expose limits for certain camera and renderer properties

tribal-tec:
This way, the UI won't be able to use a slider component, just a number edit.


plugins/RocketsPlugin/jsonPropertyMap.h
```diff
@@ -91,16 +91,28 @@ void _addPropertySchema(const PropertyMap::Property& prop,
...
-            jsonSchema["minimum"] = prop.min<T>();
-        else
-            jsonSchema.AddMember(StringRef("minimum"), prop.min<T>(),
-                                 allocator);
-        if (jsonSchema.HasMember("maximum"))
-            jsonSchema["maximum"] = prop.max<T>();
+        const auto minValue = prop.min<T>();
+        const auto maxValue = prop.max<T>();
+        const bool hasLimits = maxValue - minValue != 0;
+        if (hasLimits)
```
hernando: Late, but why not
  if (maxValue <= minValue)

PR #524, opened 2018-08-17
Make RocketsPlugin and DeflectPlugin separate libraries

tribal-tec:
* RocketsPlugin is loaded only when --http-server is specified
* DeflectPlugin is loaded via --plugin DeflectPlugin, not by default anymore


brayns/Brayns.cpp
```diff
@@ -137,6 +138,18 @@ struct Brayns::Impl : public PluginAPI
             try
             {
                 const auto& pluginName = pluginParam.name;
+
+#ifdef BRAYNS_USE_DEFLECT
+                // Deflect plugin is a built-in plugin as it still needs access
+                // to parts of the engine which are not yet exposed in the
+                // PluginAPI.
+                if (pluginName == "DeflectPlugin")
```
rdumusc: why not pluginName == "deflect" ?

tribal-tec: libdeflect.so is a different library, not the brayns plugin

rdumusc: And it has to be the library name..? It's just that it's not super intuitive to write "--plugin DeflectPlugin" with all the camel case on the command line; it also doesn't match with the --http-server. I think it would be nice to have an easy shortcut like --streaming or equivalent.

tribal-tec: For 'normal' plugins it's always the library name. Considering that, only the RocketsPlugin 'misbehaves', but this is almost a core feature rather than a real plugin. The --streaming would not decouple anything from core vs plugins, hence the deflect part shouldn't be a plugin in the first place. As one of the next steps, the DeflectPlugin should be a real plugin that is no longer linked to core, but loaded at runtime when specified.

karjonas:
lgtm

brayns/Brayns.cpp
tribal-tec: Changed the behavior now: DeflectPlugin is loaded when --stream-host or DEFLECT_HOST env is specified

rdumusc: As discussed offline, sounds good and does not break compatibility with existing launch presets

karjonas:
I think we should throw an exception if you do not have networking enabled but still provide a --http-server command line argument.

tribal-tec:
Updated

karjonas:
lgtm

PR #526, opened 2018-08-20
Add suggestions for unrecognized command line options.

karjonas:
Example:

```
[ERROR] Unrecognized option '--mode'.

Most similar options are:
	--module
	--synchronous-mode
	--circuit-uses-simulation-model
	--memory-mode
```

### DIFF
--- a/brayns/parameters/ParametersManager.cpp
+++ b/brayns/parameters/ParametersManager.cpp
@@ -24,6 +24,73 @@
 
 #include <boost/algorithm/string.hpp>
 
+namespace
+{
+// https://en.wikibooks.org/wiki/Algorithm_Implementation/Strings/Levenshtein_distance#C++
+unsigned int levenshtein_distance(const std::string& s1, const std::string& s2)
+{
+    const std::size_t len1 = s1.size(), len2 = s2.size();
+    std::vector<unsigned int> col(len2 + 1), prevCol(len2 + 1);
+
+    for (unsigned int i = 0; i < prevCol.size(); i++)
+        prevCol[i] = i;
+    for (unsigned int i = 0; i < len1; i++)
+    {
+        col[0] = i + 1;
+        for (unsigned int j = 0; j < len2; j++)
+            // note that std::min({arg1, arg2, arg3}) works only in C++11,
+            // for C++98 use std::min(std::min(arg1, arg2), arg3)
+            col[j + 1] = std::min({prevCol[1 + j] + 1, col[j] + 1,
+                                   prevCol[j] + (s1[i] == s2[j] ? 0 : 1)});
+        col.swap(prevCol);
+    }
+    return prevCol[len2];
+}
+
+std::vector<std::string> findSimilarOptions(
+    const std::string& name, const std::vector<std::string>& options)
+{
+    size_t bestDist = UINT_MAX;
+    std::vector<std::string> bestOptions;
+    constexpr size_t MAX_SUGGESTIONS = 7;
+
+    for (const auto& optionName : options)
+    {
+        const auto dist = levenshtein_distance(name, optionName);
+        if (dist < bestDist)
+        {
+            bestOptions.clear();
+            bestDist = dist;
+        }
+
+        if (dist == bestDist && bestOptions.size() < MAX_SUGGESTIONS)
+            bestOptions.push_back(optionName);
+    }
+
+    // Strip dashes
+    auto nameStrip = name;
+    nameStrip.erase(std::remove(nameStrip.begin(), nameStrip.end(), '-'),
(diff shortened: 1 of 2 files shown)


brayns/parameters/ParametersManager.cpp
```diff
@@ -24,6 +24,73 @@
...
+        if (dist == bestDist && bestOptions.size() < MAX_SUGGESTIONS)
+            bestOptions.push_back(optionName);
+    }
+
+    // Strip dashes
+    auto nameStrip = name;
+    nameStrip.erase(std::remove(nameStrip.begin(), nameStrip.end(), '-'),
+                    nameStrip.end());
+
+    // Also suggest options containing the substring
```
hernando: Maybe these suggestions should go first

brayns/parameters/ParametersManager.cpp
```diff
@@ -65,10 +132,37 @@ void ParametersManager::parse(int argc, const char** argv)
...
+
+            std::string errorMessage = "Unrecognized option '" + unrecognized +
+                                       "'.\n\nMost similar options are:";
+
+            for (const auto& suggestion : suggestions)
+                errorMessage += "\n\t" + suggestion;
+
+            throw po::error(errorMessage);
+        }
+
```
hernando: Move to a separate _processUnrecognizedOptions function?

brayns/parameters/ParametersManager.cpp
```diff
@@ -24,6 +24,73 @@
...
+    std::vector<unsigned int> col(len2 + 1), prevCol(len2 + 1);
+
+    for (unsigned int i = 0; i < prevCol.size(); i++)
+        prevCol[i] = i;
+    for (unsigned int i = 0; i < len1; i++)
+    {
+        col[0] = i + 1;
+        for (unsigned int j = 0; j < len2; j++)
+            // note that std::min({arg1, arg2, arg3}) works only in C++11,
+            // for C++98 use std::min(std::min(arg1, arg2), arg3)
```
hernando: Remove comment

brayns/parameters/ParametersManager.cpp
```diff
@@ -50,44 +48,59 @@ unsigned int levenshtein_distance(const std::string& s1, const std::string& s2)
...
+                          optionName) == subStringOptions.end())
+                levenshteinOptions.push_back(optionName);
+        }
     }
 
-    return bestOptions;
+    // Merge suggestions giving precedence to substrings
+    auto output = subStringOptions;
+    output.insert(std::end(output), std::begin(levenshteinOptions),
+                  std::end(levenshteinOptions));
```
hernando: Just a comment, why do you use std::end(..) instead of calling foo.end()? I would understand it in a template function, but here it makes the code less readable. I'd even say the same for using insert.

Don't you find this more readable and shorter to type:

    for (const auto& option : levenshteinOptions)
        output.push_back(option)

brayns/parameters/ParametersManager.cpp
```diff
@@ -50,44 +48,59 @@ unsigned int levenshtein_distance(const std::string& s1, const std::string& s2)
 std::vector<std::string> findSimilarOptions(
     const std::string& name, const std::vector<std::string>& options)
 {
-    size_t bestDist = UINT_MAX;
-    std::vector<std::string> bestOptions;
     constexpr size_t MAX_SUGGESTIONS = 7;
 
-    for (const auto& optionName : options)
+    std::vector<std::string> subStringOptions;
+    std::vector<std::string> levenshteinOptions;
```
hernando: Why two vectors that need concatenation later on? The original code was fine, I was just suggesting to move the substrings options up.

brayns/parameters/ParametersManager.cpp
```diff
@@ -258,4 +253,28 @@ void ParametersManager::set(const std::string& key, const std::string& value)
 
     parse(argc, argv.get());
 }
+
+void ParametersManager::_processUnrecognizedOptions(
+    const std::vector<std::string>& unrecognizedOptions) const
+{
+    if (!unrecognizedOptions.empty())
```
hernando: if (unrecognizedOptions.empty())
        return;

brayns/parameters/ParametersManager.cpp
karjonas: I think it is easier to just keep them separated since then I can clear just the levenshtein vector in the loop.

brayns/parameters/ParametersManager.cpp
karjonas: Yes, that is better. I would like to see a 'vector += other_vector' or 'vector.append(other_vector)' in the standard library.

PR #527, opened 2018-08-20
Add --circuit-random-seed command line flag.

karjonas:
This makes the circuit loading deterministic by default. To get randomness --circuit-random-seed is used. Using BRAIN_CIRCUIT_SEED is no longer supported.


ppodhajski:
retest this please

brayns/parameters/GeometryParameters.cpp
```diff
@@ -152,6 +153,8 @@ GeometryParameters::GeometryParameters()
         "Minimum and maximum values for the simulation [float float]")(
         PARAM_CIRCUIT_SIMULATION_HISTOGRAM_SIZE.c_str(), po::value<size_t>(),
         "Number of values defining the simulation histogram [int]")(
+        PARAM_CIRCUIT_RANDOM_SEED.c_str(), po::value<size_t>(),
```
hernando: I'd use const char* for all these strings if you really need the C representation in the end

karjonas: Seems reasonable, but that is for another PR.

karjonas:
retest this please

PR #528, opened 2018-08-21
Precision must be double for JSON

tribal-tec:


engines/ospray/utils.cpp
```diff
@@ -33,6 +33,15 @@ namespace brayns
     ospSet##OSP(ospObject, prop->name.c_str(),              \
                 prop->get<std::array<TYPE, NUM>>().data()); \
     break;
+#define SET_ARRAY_FLOAT(OSP, NUM)                                \
```
karjonas: I think we could make these defines templated functions. For another PR though.

PR #530, opened 2018-08-22
Fix ispc 32-bit address limits in SDF geometries.

karjonas:


favreau:
Ultra super cool PR. Looking at it now :)

hernando:
As discussed offline with @karjonas , the actual was problem seems to be the overflow in multiplying two int variables to obtain an offset, obviously the result becomes negative and thus the segfault. The PR could be greatly simplified by spotting the offending arithmetic and promoting the types to uint64

karjonas:
I have done some more tests now and it seems like not even ```*(buffer + index)``` is good enough. Anything using pointer arithmetics is giving issues so what I ended up doing is casting the pointer to a uint64 and doing all arithmetics and then casting it back to a pointer of the right type. See __define_safe_ptr_inc macro in ExtendedSDFGeometries.ispc.

PR #532, opened 2018-08-24
 Fix ispc 32-bit address limits in all geometries.

karjonas:


engines/ospray/ispc/geometry/ExtendedCylinders.cpp
```diff
@@ -27,43 +27,32 @@
...
+    const bool useSafeIndex = data->numBytes >= INT_MAX;
+    const size_t numExtendedCylinders = data->numBytes / bytesPerCylinder;
     ispc::ExtendedCylindersGeometry_set(getIE(), model->getIE(), data->data,
-                                        numExtendedCylinders, bytesPerCylinder,
-                                        radius, materialID, offset_center,
-                                        offset_up, offset_radius,
-                                        offset_timestamp, offset_texture_coords,
-                                        offset_materialID);
+                                        numExtendedCylinders, materialID,
+                                        useSafeIndex);
```
hernando: You can also remove useSafeIndex here

engines/ospray/ispc/geometry/ExtendedSDFGeometries.ispc
```diff
@@ -161,28 +163,9 @@ struct ExtendedSDFGeometries
...
-                                               const varying uint64 index)  \
-    {                                                                       \
-        const uniform uint64 tSize = sizeof(uniform T);                     \
-        const varying uint64 offsetBytes = tSize * index;                   \
-        return (const T* varying const)((uint64)base + offsetBytes);        \
-    }
-
 __define_safe_index(ExtendedSDFGeometry);
 __define_safe_index(uint64);
 __define_safe_index(int64);
```
hernando: You should move the defines for uint64 and int64 somewhere else as they are generic types.

engines/ospray/ispc/geometry/ExtendedSDFGeometries.cpp
```diff
@@ -82,7 +87,7 @@ void ExtendedSDFGeometries::finalize(ospray::Model* model)
                                             ispcMaterialList,
                                             numExtendedSDFGeometries,
                                             materialID, neighbours->data,
-                                            geometries->data);
+                                            geometries->data, useSafeIndex);
```
hernando: I'm not sure whether you can remove useSafeIndex from here, but give it a try.

engines/ospray/ispc/geometry/ExtendedCones.cpp
```diff
@@ -28,46 +28,29 @@
...
 ExtendedCones::ExtendedCones()
 {
     this->ispcEquivalent = ispc::ExtendedCones_create(this);
 }
 
-void ExtendedCones::finalize(ospray::Model *model)
+void ExtendedCones::finalize(ospray::Model* model)
 {
-    radius = getParam1f("radius", 0.01f);
-    length = getParam1f("length", 0.01f);
```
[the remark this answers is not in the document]
karjonas: The offsets are never set from outside this method, and radius and length were never used afaik.

engines/ospray/ispc/geometry/ExtendedCones.ispc
```diff
@@ -32,70 +32,68 @@
...
-    if (geometry->offset_centerRadius >= 0)
-        radius0 = *((uniform float *)(conePtr + geometry->offset_centerRadius));
+    uniform ExtendedCone* uniform conePtr =
+        safe_index(geometry->useSafeIndex, geometry->data, primID);
 
-    uniform float radius1 = geometry->radius;
-    if (geometry->offset_upRadius >= 0)
-        radius1 = *((uniform float *)(conePtr + geometry->offset_upRadius));
+    uniform float radius0 = conePtr->centerRadius;
+    uniform float radius1 = conePtr->upRadius;
```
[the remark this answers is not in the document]
karjonas: I think the performance is very similar since the only difference is the radius is read from the cone instead of the global geometry. Also, the whole cone geometry is always uploaded anyways so the same memory is used.

engines/ospray/ispc/geometry/ExtendedCones.ispc
```diff
@@ -250,53 +248,39 @@ static void ExtendedCones_postIntersect(uniform Geometry *uniform geometry,
         if (dot(ray.dir, Ns) >= 0.f)
             Ns = neg(Ns);
     }
-    if ((flags & DG_MATERIALID) && (this->offset_materialID >= 0))
-        dg.materialID =
-            *((uniform uint32 * varying)(conePtr + this->offset_materialID));
+    if ((flags & DG_MATERIALID) && conePtr->materialID >= 0)
+        dg.materialID = conePtr->materialID;
```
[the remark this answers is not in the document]
karjonas: It is not exactly equivalent but this code makes it possible to specify for each cone a specific material it should use which was the intention of the offset_materialID variable.

engines/ospray/ispc/geometry/ExtendedSDFGeometries.ispc
karjonas: I actually only need the uint64 type so I will remove the others and keep them here.

engines/ospray/ispc/geometry/ExtendedCones.ispc
```diff
@@ -32,70 +32,68 @@
...
+
+struct ExtendedCone
+{
+    vec3f center;
+    vec3f up;
+    float centerRadius;
+    float upRadius;
+    float timestamp;
+    vec2f texture_coords;
+    int32 materialID;
```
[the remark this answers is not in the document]
karjonas: No, I am not sure of that but ispc is supposed to interop with C/C++ and it seems to work for the cases I tried.

engines/ospray/ispc/geometry/ExtendedCones.cpp
```diff
@@ -28,46 +28,29 @@
...
-        getParam1i("offset_centerRadius",
-                   offsetof(struct brayns::Cone, centerRadius));
-    offset_upRadius =
-        getParam1i("offset_upRadius", offsetof(struct brayns::Cone, upRadius));
-    offset_timestamp = getParam1i("offset_timestamp",
-                                  offsetof(struct brayns::Cone, timestamp));
-    offset_texture_coords =
-        getParam1i("offset_texture_coords",
-                   offsetof(struct brayns::Cone, texture_coords));
-    offset_materialID = getParam1i("offset_materialID", -1);
```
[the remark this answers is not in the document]
karjonas: I have sidestepped this issue by using the same struct for spheres, cylinders and cones in both C++ and ispc.

engines/ospray/ispc/geometry/ExtendedCones.ispc
```diff
@@ -250,53 +248,39 @@ static void ExtendedCones_postIntersect(uniform Geometry *uniform geometry,
...
-    float uniform radius, float uniform length, int uniform materialID,
-    int uniform offset_center, int uniform offset_up,
-    int uniform offset_centerRadius, int uniform offset_upRadius,
-    int uniform offset_timestamp, int uniform offset_texture_coords,
-    int uniform offset_materialID)
+export void ExtendedConesGeometry_set(void* uniform _geom, void* uniform _model,
+                                      void* uniform data,
+                                      int uniform numExtendedCones,
+                                      int uniform materialID,
+                                      bool uniform useSafeIndex)
```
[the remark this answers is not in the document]
karjonas: done

karjonas:
I think I fixed the relevant issues now.

engines/ospray/ispc/geometry/utils/SafeIndex.ih
```diff
@@ -24,27 +24,36 @@
...
+        {                                                                     \
+            return base + index;                                              \
+        }                                                                     \
+    }
+
+#define __define_needs_safe_index(T)                                         \
+    inline uniform bool needs_safe_index(const T* uniform ptr,               \
+                                         const uniform uint64 numGeometries) \
+    {                                                                        \
+        return (sizeof(uniform T) * numGeometries) >= INT_MAX;               \
```
hernando: Following the code conventions, these two functions should be needsSafeIndex and safeIndex (which I would rename to safeDereference, but you may keep safeIndex). And the macro should be DEFINE_SAFE_INDEX (the double score makes it look like a compiler internal)

brayns/common/geometry/Cone.h
```diff
@@ -18,38 +18,26 @@
...
-        : center{c}
-        , up{u}
-        , centerRadius{cr}
-        , upRadius{ur}
-        , timestamp{ts}
-        , texture_coords{t}
-        , materialID(-1)
+    Cone(const Vector3f c = {0.f, 0.f, 0.f}, const Vector3f u = {0.f, 0.f, 0.f},
+         const float cr = 0.f, const float ur = 0.f, const float ts = 0.f,
+         const Vector2f t = Vector2f(0.f, 0.f), int32 m = -1)
```
hernando: Thanks to  aggregate initialization this constructor is not needed in C++11 except for providing default values (which you could do in the struct declaration, but I guess it wouldn't work for ISPC). Consider removing it.

If you remove it, I would rename ExtendedFoo to Foo. I don't see why we need Extended if there isn't a basic counterpart.

engines/ospray/ispc/geometry/utils/SafeIndex.ih
karjonas: Maybe it should be called safeIncrement since that is actually what it is doing, not dereferencing nor indexing.

karjonas:
I think I fixed the relevant issues now.

PR #533, opened 2018-08-27
Immediate notify changes on the websocket

tribal-tec:


plugins/RocketsPlugin/Timeout.h
```diff
@@ -0,0 +1,77 @@
...
+#include <condition_variable>
+#include <functional>
+#include <future>
+#include <mutex>
+
+namespace brayns
+{
+/**
+ * Implements the setTimeout() and clearTimeout() mechanics from Javascript
+ * (https://www.w3schools.com/jsref/met_win_settimeout.asp), following the same
```
rolandjitsu: [setTimeout()](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout) is a better link to the doc 😄

tests/throttle.cpp
```diff
@@ -0,0 +1,113 @@
...
+{
+    brayns::Timer timer;
+    timer.start();
+
+    brayns::Timeout timeout;
+
+    bool called = false;
+    timeout.set([&called] { called = true; }, 10000);
+    timeout.clear();
+    BOOST_CHECK(!called);
```
rolandjitsu: Shouldn't the check occur after the time passed?

1. Schedule a callback in 10 seconds
2. clear it
3. Wait for more than 10 sec
4. Check that the callback was not called

Or is the boost check covering that?

plugins/RocketsPlugin/Timeout.h
tribal-tec: Ok, thanks :smile:

tests/throttle.cpp
tribal-tec: Will add this afterwards. Not 10 seconds then of course

PR #537, opened 2018-08-29
Fix a couple of data races

tribal-tec:


plugins/RocketsPlugin/RocketsPlugin.cpp
```diff
@@ -301,6 +307,33 @@ class RocketsPlugin::Impl
...
+        std::lock_guard<std::mutex> lock(_delayedNotifiesMutex);
+
+#ifdef BRAYNS_USE_LIBUV
+        if (uvw::Loop::getDefault()->alive())
+#else
+        if (false)
+#endif
+        {
+#ifdef BRAYNS_USE_LIBUV
+            // dispatch delayed notify from throttle thread to main thread
```
karjonas: I think it is more readable to make this one ifdef/else block.

tests/addModelFromBlob.cpp
```diff
@@ -236,18 +227,15 @@ BOOST_AUTO_TEST_CASE(send_wrong_number_of_bytes)
...
     const std::string wrong("not_four_bytes");
     getWsClient().sendBinary(wrong.data(), wrong.size());
 
     try
     {
-        asyncWait.get();
+        while (!request.is_ready())
+            process();
+        request.get();
+        ;
```
karjonas: remove

plugins/RocketsPlugin/RocketsPlugin.cpp
tribal-tec: yeah, agreed

tests/addModelFromBlob.cpp
tribal-tec: done

PR #542, opened 2018-08-31
Remove runtime data loading from geometry parameters

tribal-tec:
Loading at startup from geometry parameters is still supported. Runtime loading
of new models is provided by addModel RPC.


ppodhajski:
retest this please

ppodhajski:
retest this please

PR #555, opened 2018-09-08
Expose frame buffer in plugin API

favreau:


favreau:
Retest this please

favreau:
This PR also exposes ImageUtils that can be useful to manipulate frame buffers from plugins.

brayns/common/CMakeLists.txt
```diff
@@ -73,13 +73,13 @@ set(BRAYNSCOMMON_PUBLIC_HEADERS
   transferFunction/TransferFunction.h
   types.h
   utils/Utils.h
+  utils/ImageUtils.h
```
tribal-tec: That most likely won't work when using plugins from an installed brayns as freeImage is leaking then. I would not recommend this, but if you insist, you have to make freeimage a public dependency of braynsCommon in its CMakeLists.txt.

favreau: Hum. good point! I think it's a nice feature to be able to be able to modify the frame buffer from the plugin. This allows easy implementation of post processing operations, like auto-color or sharpen for example. I would still allow the feature, but we can definitely ask the team how they feel about this?

rolandjitsu: Makes sense to be able to do post processing from the plugin, but probably this should be handled in some new callback which the plugin can register a handler for and brayns calls internally.

favreau: That sounds a bit more restrictive to me. I don't find it shocking to expose the frame buffer, since all other objects (render, camera, etc) are already exposed. Why would the frame buffer be an exception there?

favreau:
In the end, I am closing the PR because what I wanted to achieve will be satisfied by https://github.com/BlueBrain/Brayns/issues/557

PR #560, opened 2018-09-13
OptiX: Fix reference counting and proper destruction of objects

karjonas:


karjonas:
@hernando I did not consider that. Since OptiX is actually an ospray module I think it makes sense to use ospray reference counting.

hernando:
Then we should use intrusive pointers for OSPray objects ;)

PR #565, opened 2018-09-14
Moved shading modes from renderers to materials

favreau:


ppodhajski:
retest this please

ppodhajski:
retest this please

ppodhajski:
retest this please

ppodhajski:
retest this please

ppodhajski:
retest this please

ppodhajski:
retest this please

ppodhajski:
retest this please

ppodhajski:
retest this please

ppodhajski:
retest this please

ppodhajski:
retest this please

ppodhajski:
retest this please

brayns/common/CMakeLists.txt
```diff
@@ -75,13 +75,13 @@ set(BRAYNSCOMMON_PUBLIC_HEADERS
   transferFunction/TransferFunction.h
   types.h
   utils/Utils.h
+  utils/ImageUtils.h
```
tribal-tec: should remain private

brayns/Brayns.cpp
```diff
@@ -1066,4 +1047,4 @@ AbstractManipulator& Brayns::getCameraManipulator()
 {
     return _impl->getCameraManipulator();
 }
-}
+} // namespace brayns
```
tribal-tec: ? different clang-format version why?

ospray_modules/optix/cuda/renderer/Phong.h
```diff
@@ -263,7 +261,8 @@ static __device__ void phongShadowed(float3 p_Ko)
 static __device__ void phongShade(float3 p_Kd, float3 p_Ka, float3 p_Ks,
                                   float3 p_Kr, float3 p_Ko,
                                   float p_refractionIndex, float p_phong_exp,
-                                  float p_glossiness, float3 p_normal,
+                                  float p_glossiness,
+                                  unsigned int p_shadindMode, float3 p_normal,
```
karjonas: typo shadin(d)Mode

brayns/Brayns.cpp
favreau: hum... yes, I am using clang-format version 3.8.0-2ubuntu4... Which one are you using?

favreau: hold on, this probably comes from my home PC where I am using Ubuntu 18.04, or maybe from my Mac... Let me put things back together from my Ubuntu machine.

engines/ospray/ispc/render/ExtendedOBJMaterial.ih
```diff
@@ -28,6 +28,14 @@
 #include <ospray/SDK/texture/TextureParam.ih>
 #include <ospray/SDK/volume/Volume.ih>
 
+// Caution: This has to match the Brayns::MaterialShadingMode
+enum MaterialShadingMode
```
karjonas: Can you not put this in a header file somewhere and include that in ispc and cpp code?

brayns/Brayns.cpp
tribal-tec: also 3.8 from Ubuntu 16.04. It must be a more recent version, yes.

engines/ospray/ispc/render/ExtendedOBJMaterial.ih
favreau: I tried, but could not find a way to do so... I am having another look now, with not much hope though.

tribal-tec:
The `CMake/common` change probably also needs to be undone, as it looks like it reverts back one commit?

engines/ospray/ispc/render/ExtendedOBJMaterial.ih
karjonas: I did something similar in brayns/common/geometry/Cone.h

brayns/common/CMakeLists.txt
```diff
@@ -81,7 +81,6 @@ set(BRAYNSCOMMON_PUBLIC_HEADERS
 )
 
 set(BRAYNSCOMMON_HEADERS
-  utils/ImageUtils.h
```
tribal-tec: undo delete?

tests/braynsTestData.cpp
```diff
@@ -119,6 +120,24 @@ BOOST_AUTO_TEST_CASE(render_circuit_with_color_and_compare)
     camera.setTarget(rotCenter);
     camera.setPosition(camPos + 0.9 * (rotCenter - camPos));
 
+    auto& scene = brayns.getEngine().getScene();
```
tribal-tec: that outlines a bit of the problem of this approach: the user now cannot 'easily' change the shading of the entire scene (in the UI not all anymore). Also, renderers which don't evaluate the material shading mode (like the basic renderer), shade anyway the way they want :)

hernando:
What are the motivations for this change? I just read the commit name, but logically it doesn't make much sense to me.

favreau:
The shading is a material property, not a renderer one. I understand this is more convenient from a user point of view to change everything in one click, but in reality, one needs more granularity. The bounding box is a good example of geometries that needs specific shading. To me, having this attribute at a renderer level is what does not make sense.

brayns/common/CMakeLists.txt
favreau: oops, thanks.

hernando:
I misread shading by shadowing. Maybe it makes sense after all.

straaljager:
"The shading is a material property, not a renderer one." I think it would be good to discuss this type of design choice. As far as I know, all rendering engines on the market use the same shading (e.g. AO, path tracing, Phong) for all materials in the scene. I'm not showing we should be doing exactly the same as what other renderers out there are doing, as there might be good reasons to implement it differently from the traditional way. But that needs discussion, as currently it's unclear.

straaljager:
Sounds like there is a bit of confusion about terminology. Disregard my previous comment.

favreau:
No worries. Just to clarify, I am not dealing with rendering options there, this PR only is about the "color" of the surface according to the light incidence. It's somewhat like an extremely simplified BRDF implementation. We will eventually have to implement proper BRDF materials, and this will of course require discussions.

straaljager:
OK, it all makes sense now. Looking forward to better materials and accompanying discussions :)

CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ cmake_minimum_required(VERSION 3.1 FATAL_ERROR)
 project(Brayns VERSION 0.8.0)
 set(Brayns_VERSION_ABI 1)
 
+set(BRAYNS_SOURCE_DIR ${PROJECT_SOURCE_DIR})
```
tribal-tec: ? Use `CMAKE_SOURCE_DIR` where you need to get this.

brayns/common/commonTypes.h
```diff
@@ -0,0 +1,20 @@
+#ifndef COMMONTYPES_H
```
tribal-tec: no copyright header here?

brayns/common/types.h
```diff
@@ -30,6 +30,7 @@
 #define SERIALIZATION_FRIEND(type) \
     friend void staticjson::init(type*, staticjson::ObjectHandler*);
 
+#include "commonTypes.h"
```
tribal-tec: should be system include like the one below

tests/braynsTestData.cpp
```diff
@@ -119,6 +120,24 @@ BOOST_AUTO_TEST_CASE(render_circuit_with_color_and_compare)
...
+    for (size_t i = 0; i < scene.getNumModels(); ++i)
+    {
+        auto modelDescriptor = scene.getModel(i);
```
tribal-tec: this is wrong btw, as `getModel` needs a modelID, not an index.

tribal-tec:
All the test images should not change, right? Maybe just revert of all them?

PR #568, opened 2018-09-19
OptiX-related fixes

karjonas:


brayns/parameters/ApplicationParameters.cpp
```diff
@@ -104,11 +104,20 @@ void ApplicationParameters::parse(const po::variables_map& vm)
 {
     if (vm.count(PARAM_ENGINE))
     {
-        _engine = EngineType::ospray;
+        bool matched = false;
         const std::string& engine = vm[PARAM_ENGINE].as<std::string>();
-        for (size_t i = 0; i < sizeof(ENGINES) / sizeof(ENGINES[0]); ++i)
+        for (size_t i = 0; i < ENGINES.size(); ++i)
```
hernando: use std::find?

ospray_modules/optix/Context.cpp
```diff
@@ -227,7 +192,7 @@ ::optix::TextureSampler Context::getTextureSampler(ospray::Texture2D* tx)
     return _optixTextureSamplers[tx];
 }
 
-void Context::updateLights(ospray::Ref<ospray::Data> lightData)
+void Context::updateLights(ospray::Data* lightData)
```
hernando: Can it be const osgpray::Data*?

ospray_modules/optix/framebuffer/FrameBuffer.cpp
```diff
@@ -117,14 +117,15 @@ float FrameBuffer::endFrame(const float /*errorThreshold*/)
 const void* FrameBuffer::mapDepthBuffer()
 {
     if (colorBuffer)
-        depthBuffer = (float*)colorBuffer;
+        depthBuffer = colorBuffer;
     return (const void*)depthBuffer;
```
hernando: I cannot judge this code without studying the whole design, but this looks strange.

ospray_modules/optix/Context.cpp
```diff
@@ -283,50 +248,155 @@ void Context::destroy()
...
+    else // Old encoding with only one digit for the micro number.
     {
-        char deviceName[256];
-        int computeCaps[2];
-        int clock_rate;
-        uint64_t totalMemory{0};
+        minor = (optixVersion % 1000) / 10;
+        micro = optixVersion % 10;
+    }
+    ospray::postStatusMsg() << "OptiX " << major << "." << minor << "."
```
hernando: Can postStatusMsg be used without initializing OSPray, or is it always done anyway?

karjonas: This is dependent on OSPRay but since the optix plugin is actually a plugin that is loaded by OSPRay it should always be set.

ospray_modules/optix/framebuffer/FrameBuffer.cpp
karjonas: I am not sure either but I think  it is like this to hide the fact that we do not provide a depth buffer from OptiX.

karjonas:
retest this please

karjonas:
I fixed the issues and added some more commits. Any more comments?

brayns/parameters/ApplicationParameters.cpp
```diff
@@ -104,11 +104,14 @@ void ApplicationParameters::parse(const po::variables_map& vm)
...
     if (vm.count(PARAM_ENGINE))
     {
-        _engine = EngineType::ospray;
         const std::string& engine = vm[PARAM_ENGINE].as<std::string>();
-        for (size_t i = 0; i < sizeof(ENGINES) / sizeof(ENGINES[0]); ++i)
-            if (engine == ENGINES[i])
-                _engine = static_cast<EngineType>(i);
+        const size_t pos =
+            std::distance(ENGINES.begin(),
+                          std::find(ENGINES.begin(), ENGINES.end(), engine));
```
hernando: Be careful with distance, it's complexity is linear even for std::vector. If you know you have a vector is much better to do iterator subtraction.

In this case, I woundn't even use it. Replace pos != ENGINES.size() with pos != ENGINES.end() and pos with *pos.

ospray_modules/optix/framebuffer/FrameBuffer.cpp
hernando: That looks totally wrong, but I guess it can go in another PR

brayns/parameters/ApplicationParameters.cpp
karjonas: I am using a map now to avoid all indexing and type casting.

PR #569, opened 2018-09-19
Fix deadlock when updating camera & renderer properties via Rockets

tribal-tec:
Due the data race fixes in the latest Rockets versions, it revealed this wrong
behavior in Brayns, where the updates on property objects from within Rockets
shall not trigger the onModified() callback which would lead to an
unnecessary (and deadlocking) notification.

### DIFF
--- a/plugins/RocketsPlugin/RocketsPlugin.cpp
+++ b/plugins/RocketsPlugin/RocketsPlugin.cpp
@@ -1075,13 +1075,17 @@ public:
 
                 if (!document.HasMember("id") ||
                     !document.HasMember("properties"))
+                {
                     return Response::invalidParams();
+                }
 
                 const auto modelID = document["id"].GetInt();
                 auto model = _engine->getScene().getModel(modelID);
                 if (!model)
+                {
                     return Response{
                         Response::Error{"Model not found", MODEL_NOT_FOUND}};
+                }
 
                 Document propertyDoc;
                 propertyDoc.SetObject() = document["properties"].GetObject();
@@ -1214,7 +1218,7 @@ public:
             PropertyMap props = object.getPropertyMap();
             if (::from_json(props, request.message))
             {
-                object.updateProperties(props);
+                object.updateProperties(props, false);
                 _engine->triggerRender();
 
                 this->_rebroadcast(notifyEndpoint, props, request.clientID);
--- a/plugins/RocketsPlugin/RocketsPlugin.cpp
+++ b/plugins/RocketsPlugin/RocketsPlugin.cpp
@@ -173,7 +173,7 @@ inline bool from_json(T& obj, const std::string& json,
         staticjson::from_json_string(json.c_str(), &obj, &status);
     if (success)
     {
-        obj.markModified(false);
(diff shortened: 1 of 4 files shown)


plugins/RocketsPlugin/RocketsPlugin.cpp
```diff
@@ -1214,7 +1218,7 @@ class RocketsPlugin::Impl
             PropertyMap props = object.getPropertyMap();
             if (::from_json(props, request.message))
             {
-                object.updateProperties(props);
+                object.updateProperties(props, false);
```
hernando: Is it OK to not update the UI if you make changes from a notebook? Or does the rebroadcast at the bottom solve that problem?

rolandjitsu: Try it out :)

tribal-tec: Yes, the rebroadcast solves that. The callback triggered by markModified() is needed for 'internal' and plugin changes.

tribal-tec:
I'll fix this differently, hang tight and hold on.

tribal-tec:
Updated. Please review again.

plugins/RocketsPlugin/RocketsPlugin.cpp
```diff
@@ -345,20 +345,97 @@ class RocketsPlugin::Impl
...
+        {
+            *_currentClientID = newID;
+        }
+
+        ~ScopedCurrentClient() { *_currentClientID = NO_CURRENT_CLIENT; }
+    private:
+        uintptr_t* _currentClientID;
+    };
+
+    void bindEndpoint(const std::string& method,
```
hernando: No leading _?

plugins/RocketsPlugin/RocketsPlugin.cpp
```diff
@@ -1254,6 +1328,9 @@ class RocketsPlugin::Impl
     std::unordered_map<std::string, std::pair<std::mutex, Throttle>> _throttle;
     std::vector<std::function<void()>> _delayedNotifies;
     std::mutex _delayedNotifiesMutex;
+    static constexpr uintptr_t NO_CURRENT_CLIENT{
+        std::numeric_limits<uintptr_t>::max()};
```
hernando: If this is representing an opaque pointer, you can also use 0 for brevity.

plugins/RocketsPlugin/RocketsPlugin.cpp
```diff
@@ -345,20 +345,97 @@ class RocketsPlugin::Impl
...
+            }
+            catch (const rockets::jsonrpc::response_error& e)
+            {
+                return Response{Response::Error{e.what(), e.code}};
+            }
+        });
+    }
+
+    template <typename Params>
+    void _connectEndpoint(const std::string& method,
```
hernando: Why these methods are not called bind as well? Because they return OK?

tribal-tec: Because in rockets they are called connect as well.

PR #570, opened 2018-09-20
Fix clear() for Timeout and improve robustness of throttle test

tribal-tec:


tests/throttle.cpp
```diff
@@ -26,32 +26,69 @@
...
-    BOOST_CHECK(!called);
+    timer.wait_for(waitTime);
 
-    timer.stop();
-    BOOST_CHECK_LE(timer.milliseconds(), 500);
+    BOOST_CHECK_LT(timer.milliseconds(), waitTime);
 
-    std::this_thread::sleep_for(std::chrono::milliseconds(2));
-    BOOST_CHECK(!called);
+    std::this_thread::sleep_for(
```
rdumusc: Once the code gets here, it should normally be safe to BOOST_CHECK(!timer.wasCalled()); right? So is this extra wait here to cover the case of spurrious wakeup in TestTimer::wait_for? If that's the case, I noticed that there is an overload of condition.wait_for() that takes a predicate (here that would be []{!called}) that takes care of removing the spurrious wakeups. It could make the timer.wait_for() call predicatable and simplify the testing.

tribal-tec: The sleep does not make sense anymore, good catch. The spurious wakeup could be an issue, I will look at this predicate for wait_for, thanks!

PR #573, opened 2018-09-21
Fix crash when not deleting the last model

tribal-tec:


brayns/common/scene/Scene.cpp
```diff
@@ -156,7 +156,7 @@ void Scene::removeModel(const size_t id)
 
         (*i)->callOnRemoved();
 
-        _modelDescriptors.erase(i, _modelDescriptors.end());
```
hernando: I had made the same change in my code because I found the usage of remove_if unnecessary and confusing. I didn't notice that it was even wrong.

Issue #516, opened 2018-08-15
Apply clip planes to other cameras


rolandjitsu:
Clip planes are currently applied only to `clippedperspective` camera. This should not be the case.

Clip planes should be applied to both the `perspective` and `orthographic` camera. And as a consequence, `clippedperspective` camera can be removed.

hernando:
For clarification, clip planes are now a scene parameter. Clipping planes being a camera property in OSPRay is just an implementation detail.

hernando:
Closed by #576

PR #576, opened 2018-09-25
Clip planes in ortho camera and new API

hernando:


brayns/common/scene/Scene.cpp
```diff
@@ -147,26 +176,17 @@ void Scene::removeModel(const size_t id)
...
-                                });
-        if (i == _modelDescriptors.end())
-            return;
-
-        (*i)->callOnRemoved();
-
-        _modelDescriptors.erase(i);
+        auto model =
+            _remove(_modelDescriptors, id, &ModelDescriptor::getModelID);
+        model->callOnRemoved();
```
karjonas: model can be null here so add null check before callOnRemoved().

brayns/common/scene/Scene.cpp
```diff
@@ -35,10 +35,39 @@ namespace fs = boost::filesystem;
 namespace
 {
 const size_t CACHE_VERSION = 10;
+
+template <typename T, typename U = T> // U seems to be needed when getID is a
+                                      // member function of a base of T.
+std::shared_ptr<T> _find(const std::vector<std::shared_ptr<T>>& list,
```
karjonas: I think it would look nicer to take a lambda instead of a getID method, so you call _find_if(list, [id](const auto& x){ return x==id;}).

brayns/common/scene/Scene.cpp
```diff
@@ -213,6 +233,26 @@ bool Scene::empty() const
     return true;
 }
 
+size_t Scene::addClipPlane(const Plane& plane)
+{
+    auto clipPlane = std::make_shared<ClipPlane>(plane);
+    clipPlane->onModified([&](const BaseObject&){ markModified(); });
+    _clipPlanes.emplace_back(std::move(clipPlane));
+    markModified();
+    return (*_clipPlanes.rbegin())->getID();
```
karjonas: _clipPlanes.back()->getID() ?

brayns/common/scene/Scene.h
```diff
@@ -30,10 +30,37 @@
...
+        _updateValue(_plane, plane);
+    }
+
+    /** @internal */
+    ClipPlane() {}
+    /** @internal */
+    ClipPlane(const size_t id, const Plane& plane) : _id(id), _plane(plane) {}
+
+protected:
+    static size_t _nextID;
```
karjonas: Should this be atomic for threading safety?

brayns/common/scene/Scene.h
```diff
@@ -30,10 +30,37 @@
 
 #include <shared_mutex>
 
+SERIALIZATION_ACCESS(ClipPlane)
 SERIALIZATION_ACCESS(Scene)
 
 namespace brayns
 {
+
+class ClipPlane : public BaseObject
```
tribal-tec: move this to a separate file

brayns/common/scene/Scene.h
```diff
@@ -30,10 +30,37 @@
...
+    size_t getID() const { return _id; }
+
+    const Plane& getPlane() const { return _plane; };
+    void setPlane(const Plane& plane)
+    {
+        _updateValue(_plane, plane);
+    }
+
+    /** @internal */
+    ClipPlane() {}
```
tribal-tec: = default

brayns/common/scene/Scene.h
```diff
@@ -152,18 +179,20 @@ class Scene : public BaseObject
         return _parametersManager;
     }
 
+    BRAYNS_API size_t addClipPlane(const Plane& plane);
```
tribal-tec: a bit of doxygen?

engines/ospray/ispc/camera/OrthographicCamera.h
```diff
@@ -0,0 +1,70 @@
...
+// limitations under the License.                                           //
+// ======================================================================== //
+
+#pragma once
+
+#include "camera/Camera.h"
+
+namespace ospray
+{
+/*! \defgroup orthographic_camera The Orthographic Camera ("orthographic")
```
tribal-tec: remove this defgroup stuff is it clutters the doxygen output

Changelog.md
```diff
@@ -592,4 +601,4 @@
 
 
 
-\* *This Change Log was automatically generated by [github_changelog_generator](https://github.com/skywinder/Github-Changelog-Generator)*
\ No newline at end of file
+\* *This Change Log was automatically generated by [github_changelog_generator](https://github.com/skywinder/Github-Changelog-Generator)*
```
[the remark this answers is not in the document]
hernando: It's difficult to see that when it's the last line of the log

Changelog.md
```diff
@@ -1,6 +1,15 @@
 # Change Log
 
-## [0.7.0](https://github.com/BlueBrain/Brayns/tree/0.7.0) (05-09-2018)
+## master
+
+**Breaking changes:**
+
+- clippedperspective camera moved to perspective [\#576](https://github.com/BlueBrain/Brayns/pull/576)
+- New scene API for add/get/update/remove clip planes. [\#576](https://github.com/BlueBrain/Brayns/pull/576)
```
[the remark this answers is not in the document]
hernando: Somehow it is. The UI for clip planes doesn't work at all after this change.

brayns/common/scene/Scene.h
hernando: I discussed it with @tribal-tec, it should be safe as only the main thread creates ClipPlanes

brayns/common/scene/Scene.cpp
hernando: true, good catch

Changelog.md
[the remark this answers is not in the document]
tribal-tec: The new functions are not, but the removal of the handling of taking the scene json into account, yes. I would then probably prefer pushing my bugfix PR first, and then you would actually have to bump the (minor) version.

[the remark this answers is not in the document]
rolandjitsu: That should not be the case, the direct scene updates should not be affected.

Changelog.md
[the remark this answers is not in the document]
tribal-tec: complain to `https://github.com/skywinder/Github-Changelog-Generator` :)

brayns/common/scene/Scene.h
hernando: Can I still keep the definition of _nextID in Scene.cpp (creating a new file just for that is annyoing)?

Changelog.md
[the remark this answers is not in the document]
rolandjitsu: We might want to use a different tool, such as [standard-version](https://github.com/conventional-changelog/standard-version) and adjust it to bump versions in the files we need.

engines/ospray/ispc/camera/OrthographicCamera.h
hernando: I reworked the whole doxygen comment a bit

brayns/common/scene/Scene.cpp
hernando: I must be missing something because what you say doesn't compile in my head.

brayns/common/scene/Scene.h
tribal-tec: Adding a file is annoying? Having something from a different file in the scene.cpp is annoying :)

tribal-tec:
Also reference the issue  #516 to close once this PR gets merged.

Changelog.md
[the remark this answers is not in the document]
hernando: I have removed clip_planes from the scene json to reflect the intent of this PR. Regretfully, the scene is still broadcasted when clip planes are modified, but that can't be solved without bigger changes and solve be addressed for models at the same time. What we have discussed offline is to remove the scene object from the API. There's a single implicit scene and RPC methods get/set its components separately.

By the way, do I need to provide a PUT end-point of updating planes?

brayns/common/scene/Scene.h
hernando: Yes it's annoying

brayns/common/scene/ClipPlane.h
```diff
@@ -0,0 +1,60 @@
+/* Copyright (c) 2015-2018, EPFL/Blue Brain Project
+ * All rights reserved. Do not distribute without permission.
+ * Responsible Author: Cyrille Favreau <cyrille.favreau@epfl.ch>
```
tribal-tec: why not you?

brayns/common/scene/ClipPlane.h
```diff
@@ -0,0 +1,60 @@
...
+    /** @internal */
+    ClipPlane() = default;
+    /** @internal */
+    ClipPlane(const size_t id, const Plane& plane)
+        : _id(id)
+        , _plane(plane)
+    {
+    }
+
+protected:
```
tribal-tec: private?

brayns/common/scene/Scene.cpp
```diff
@@ -213,6 +235,26 @@ bool Scene::empty() const
...
+
+ClipPlanePtr Scene::getClipPlane(const size_t id) const
+{
+    return _find(_clipPlanes, id);
+}
+
+void Scene::removeClipPlane(const size_t id)
+{
+    _remove(_clipPlanes, id);
+    markModified();
```
tribal-tec: `if(_remove())` to not trigger unnecessary scene updates?

brayns/common/scene/Scene.cpp
karjonas: I meant like this: https://gist.github.com/karjonas/b59a568c07f26acafa5f2541557292f8

hernando: It's more generic, but I don't find it nicer because the client code is longer. I'd change it only if the function needs to be more generic.

ppodhajski:
retest this please

hernando:
CI is finally passing, any further comments?

rolandjitsu:
@hernando I will take a look as well tomorrow morning, so hold on with merging.

Changelog.md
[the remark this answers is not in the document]
rolandjitsu: Probably this file should have no changes.

brayns/common/scene/Scene.cpp
```diff
@@ -145,28 +173,25 @@ size_t Scene::addModel(ModelDescriptorPtr model)
 
 void Scene::removeModel(const size_t id)
 {
+    bool removed = false;
```
rolandjitsu: Is this a fix some other issue? Or is this required for the clip planes?

brayns/common/types.h
```diff
@@ -79,27 +79,27 @@ class Brayns;
 class ActionInterface;
 
 class Engine;
-typedef std::shared_ptr<Engine> EnginePtr;
+using EnginePtr = std::shared_ptr<Engine>;
```
rolandjitsu: These changes could have been part of a different PR as it's a refactor of non clip plane specific code.

brayns/common/types.h
```diff
@@ -297,8 +301,8 @@ enum class CameraMode
  * in absolute value of the coordinate system. Values are stored
  * in a Vector4, with the following order: nx, ny, nz and d
  */
-using ClipPlane = std::array<double, 4>;
-using ClipPlanes = std::vector<ClipPlane>;
+using Plane = std::array<double, 4>;
```
rolandjitsu: Why the rename? `ClipPlane` still makes more sense.

tests/snapshot.cpp
```diff
@@ -44,8 +44,8 @@ BOOST_AUTO_TEST_CASE(snapshot)
 {
     brayns::SnapshotParams params;
     params.format = "png";
-    params.size = {50, 50};
```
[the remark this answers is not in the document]
hernando: Because I wanted to reuse the image for the clipping plane tests. I can make all test images 50 x 50 as well. For my display 50 by 50 was too small, but for the test probably doesn't matter

brayns/common/scene/Scene.cpp
hernando: No, but Daniel requested not to call markModified if nothing changed for clip planes, so I propagated the change here.

Changelog.md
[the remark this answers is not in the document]
hernando: I don't spot the change, maybe a space at the end of line? I've reverted the file anyway

brayns/common/types.h
hernando: That's called continuous refactoring :)

brayns/common/types.h
hernando: Because ClipPlane is now a different type (a Plane + an ID)

brayns/common/scene/Scene.h
hernando: No, I don't check the coefficients

PR #579, opened 2018-09-26
Release 0.7.1

tribal-tec:


rolandjitsu:
@ppodhajski can we have the bbpbot **not** post these redundant comments? The checks already show the status of the build and you can always click details to see more about it.

And for more sophisticated checks details, see [checks-api](https://blog.github.com/2018-05-07-introducing-checks-api/).

hernando:
I second Roland's request, because the messages are misleading. I don't know what they are supposed to be reporting, but CI hasn't finished by the time the message gets posted, and it can still fail indeed.

PR #586, opened 2018-10-10
Use loader registry for all loaders

karjonas:
This commit unifies the code path for all loading. It also adds support for specifying the name of a certain loader to be used.


brayns/Brayns.cpp
```diff
@@ -426,159 +369,114 @@ struct Brayns::Impl : public PluginAPI
...
+                task.result();
+            }
+        }
 
-            const std::string& cacheFile(geometryParameters.getNESTCacheFile());
-            if (!geometryParameters.getNESTReport().empty() &&
-                cacheFile.empty())
-                throw std::runtime_error(
-                    "Need cache file to visualize simulation data");
+        if (!geometryParameters.getCircuitConfiguration().empty())
```
tribal-tec: Let's remove this and the `getCircuitConfiguration()`; circuits can already be loaded by regular `--input-paths`

brayns/common/loader/LoaderRegistry.h
```diff
@@ -35,11 +35,14 @@ class LoaderRegistry
...
+        std::string name;
+
         /**
-         * The function that returns the supported types (extension, filename
-         * patterns) of the loader.
+         * The set of supported types (extension, filename patterns) of the
+         * loader.
          */
-        std::function<std::set<std::string>()> supportedTypes;
+        std::set<std::string> supportedTypes;
```
hernando: Since you're changing this, I think it's better to register a function that returns true if a file is supported and false otherwise instead of the set of extensions (types should also be renamed to extension or something closer to what is actually means). The reason is that SONATA circuits and simulations use both `.json`, but the files are different. We may consider having a SimulationLoader and a CircuitLoader, but even if we don't, `.json` is too much of a generic extension to just assume that a single loader will load everything.

brayns/common/scene/Model.h
```diff
@@ -96,9 +96,15 @@ class ModelParams : public ModelInstance
     const std::string& getName() const { return _name; }
     void setPath(const std::string& path) { _updateValue(_path, path); }
     const std::string& getPath() const { return _path; }
+    void setLoaderName(const std::string& loaderName)
+    {
+        _updateValue(_loaderName, loaderName);
+    }
+    const std::string& getLoaderName() const { return _loaderName; }
```
hernando: I think this wouldn't be needed following my previous suggestion. Eventually, GeometryParams is going to disappear and you'll have to revisit this change because the core won't have enough information to decide the loader name by itself, and in any case, it will depend in the plugins that are loaded.

brayns/io/MolecularSystemReader.cpp
```diff
@@ -38,6 +38,10 @@ MolecularSystemReader::MolecularSystemReader(
     , _geometryParameters(geometryParameters)
 {
 }
+std::set<std::string> MolecularSystemReader::getSupportedDataTypes()
+{
+    return {};
```
hernando: Empty?
I'm looking at the commits one at a time, so maybe it was addressed later.

brayns/Brayns.cpp
```diff
@@ -468,62 +479,11 @@ struct Brayns::Impl : public PluginAPI
...
         scene.buildEnvironmentMap();
         scene.markModified();
     }
 
-#if (BRAYNS_USE_BRION)
-    /**
-        Loads morphologies from circuit configuration (command line
-       parameter --circuit-configuration)
-    */
-    void _loadCircuitConfiguration(const Loader::UpdateCallback& progressUpdate)
```
hernando: If you remove the progress update (which I guess you do because it doesn't show in the UI anyway), you should mention it in the commit message as well.

brayns/Brayns.cpp
```diff
@@ -376,41 +376,44 @@ struct Brayns::Impl : public PluginAPI
         auto& geometryParameters = _parametersManager.getGeometryParameters();
 
         registry.registerLoader(
-            {"MeshLoader", MeshLoader::getSupportedDataTypes(), [&] {
+            {MeshLoader::getName(), MeshLoader::getSupportedDataTypes(), [&] {
                  return std::make_unique<MeshLoader>(scene, geometryParameters);
              }});
```
hernando: This is so repetitive that it makes sense to add a macro back, but one that reads

REGISTER_LOADER(Foo, ...)

A draft would be:

    #define REGISTER_LOADER(ClassName) \
        registry.registerLoader({#ClassName, ClassName::getSupportedDataTypes(), [&](){ \
                 return std::make_unique<ClassName>(scene, __VA_ARGS__); }})

You don't need getName with this macro

karjonas:
I ended up changing so much code I decided to rename this PR and squash the changes.

brayns/io/CircuitLoader.cpp
```diff
@@ -365,15 +367,16 @@ class CircuitLoader::Impl
...
+                                      transformation, materialId);
             }
             catch (...)
             {
                 ++loadingFailures;
             }
             ++meshIndex;
-            _parent.updateProgress(message.str(), meshIndex, gids.size());
+            callback.updateProgress(message.str(),
+                                    meshIndex / (1.f * gids.size()));
```
tribal-tec: `1.f * size`? convert to float instead?

brayns/io/CircuitLoader.cpp
```diff
@@ -479,31 +484,49 @@ CircuitLoader::~CircuitLoader()
 {
 }
 
-std::set<std::string> CircuitLoader::getSupportedDataTypes()
+bool CircuitLoader::isSupported(const std::string& filename,
+                                const std::string& extension
+                                    BRAYNS_UNUSED) const
 {
-    return {"BlueConfig", "BlueConfig3", "CircuitConfig", "circuit"};
+    const auto ends_with = [](std::string const& value,
```
tribal-tec: right-hand side const? Not consistent with the rest of the code.

brayns/io/MolecularSystemReader.cpp
```diff
@@ -103,19 +115,22 @@ bool MolecularSystemReader::_createScene()
...
+                                          proteinCount, materialId);
 
-                if (_proteinFolder.empty())
+                if (data._proteinFolder.empty())
                     ++proteinCount;
             }
 
-        updateProgress("Loading proteins...", proteinCount, _nbProteins);
+        data._callback.updateProgress("Loading proteins...",
+                                      proteinCount / (1.f * data._nbProteins));
```
tribal-tec: `float(data._nbProteins)`

brayns/io/ProteinLoader.h
```diff
@@ -36,14 +36,16 @@ class ProteinLoader : public Loader
 public:
     ProteinLoader(Scene& scene, const GeometryParameters& geometryParameters);
 
-    static std::set<std::string> getSupportedDataTypes();
-
+    virtual bool isSupported(const std::string& filename,
```
tribal-tec: rm those bloody virtuals

tests/addModel.cpp
```diff
@@ -175,13 +172,6 @@ BOOST_AUTO_TEST_CASE(close_client_while_pending_request)
     BOOST_CHECK_THROW(asyncWait.get(), rockets::jsonrpc::response_error);
 }
 
-BOOST_AUTO_TEST_CASE(folder)
```
tribal-tec: keep the test, now it should throw.

karjonas:
Will add 'Remove folder loading' when I merge.

hernando:
I haven't got into the details but it seems you addressed my concerns and did a pretty nice clean up. Good job!

PR #589, opened 2018-10-12
Moving to OSPRay 1.7.x

hernando:
Dropped support for previous OSPRay versions

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,15 +113,11 @@ endif()
 list(APPEND LCOV_EXCLUDE '${PROJECT_SOURCE_DIR}/perceptualdiff/*')
 
 # OSPRay rendering engine
-common_find_package(ospray 1.5 SYSTEM)
-set(_ospray_unsupported_version 1.6)
-if(OSPRAY_FOUND AND OSPRAY_VERSION VERSION_LESS _ospray_unsupported_version)
+common_find_package(ospray 1.7 SYSTEM)
+if(OSPRAY_FOUND)
   option(BRAYNS_OSPRAY_ENABLED "Activate OSPRay rendering engine" ON)
 else()
   unset(BRAYNS_OSPRAY_ENABLED)
-  if(OSPRAY_FOUND)
-    message(WARNING "OSPRay version >= ${_ospray_unsupported_version} not supported")
-  endif()
 endif()
 
 # Mesh loading
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,9 +52,8 @@ ENV PATH $PATH:${ISPC_PATH}
 
 # Install embree
 # https://github.com/embree/embree/releases
-ARG EMBREE_VERSION=2.17.4
+ARG EMBREE_VERSION=3.2.3
 ARG EMBREE_FILE=embree-${EMBREE_VERSION}.x86_64.linux.tar.gz
-
 RUN mkdir -p ${DIST_PATH} \
   && wget https://github.com/embree/embree/releases/download/v${EMBREE_VERSION}/${EMBREE_FILE} \
   && tar zxvf ${EMBREE_FILE} -C ${DIST_PATH} --strip-components=1 \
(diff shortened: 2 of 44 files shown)


hernando:
In order to pass CI here we need to update the nix branch, let me know if you're OK with doing it now.

CMakeLists.txt
```diff
@@ -105,8 +105,8 @@ endif()
 list(APPEND LCOV_EXCLUDE '${PROJECT_SOURCE_DIR}/perceptualdiff/*')
 
 # OSPRay rendering engine
-common_find_package(ospray 1.5 SYSTEM)
-set(_ospray_unsupported_version 1.6)
+common_find_package(ospray 1.7 SYSTEM)
+set(_ospray_unsupported_version 1.8)
```
tribal-tec: this doesn't make sense as we don't know yet if it might work with 1.8 or not. Just remove the usage of that `_ospray_unsupported_version`

Dockerfile
```diff
@@ -52,23 +52,22 @@ ENV PATH $PATH:${ISPC_PATH}
...
-
 RUN mkdir -p ${DIST_PATH} \
   && wget https://github.com/embree/embree/releases/download/v${EMBREE_VERSION}/${EMBREE_FILE} \
   && tar zxvf ${EMBREE_FILE} -C ${DIST_PATH} --strip-components=1 \
   && rm -rf ${DIST_PATH}/bin ${DIST_PATH}/doc
 
 # Install OSPRay
 # https://github.com/ospray/ospray/releases
-ARG OSPRAY_VERSION=1.5.0-patch
+ARG OSPRAY_BRANCH=release-1.7.x
```
tribal-tec: why not using the tag?

engines/ospray/OSPRayMaterial.cpp
```diff
@@ -43,7 +45,8 @@ static TextureTypeMaterialAttribute textureTypeMaterialAttribute[8] = {
     {TT_REFRACTION, "map_refraction"}};
 
 OSPRayMaterial::OSPRayMaterial()
-    : _ospMaterial(ospNewMaterial(nullptr, "ExtendedOBJMaterial"))
+    // No renderer specific materials for the moment
```
tribal-tec: same status as before, so no reason for that comment

engines/ospray/OSPRayMaterial.cpp
```diff
@@ -118,12 +121,19 @@ OSPTexture2D OSPRayMaterial::_createOSPTexture2D(Texture2DPtr texture)
                  << ": " << texture->getWidth() << "x" << texture->getHeight()
                  << "x" << (int)type << std::endl;
 
-    osp::vec2i texSize{int(texture->getWidth()), int(texture->getHeight())};
-    OSPTexture2D ospTexture =
-        ospNewTexture2D(texSize, type, texture->getRawData(),
-                        OSP_TEXTURE_SHARED_BUFFER);
+    OSPTexture ospTexture = ospNewTexture("texture2d");
+
+    osp::vec2i size{int(texture->getWidth()), int(texture->getHeight())};
```
tribal-tec: const?

engines/ospray/OSPRayMaterial.cpp
```diff
@@ -118,12 +121,19 @@ OSPTexture2D OSPRayMaterial::_createOSPTexture2D(Texture2DPtr texture)
...
 
-    assert(ospTexture);
+    ospSet1i(ospTexture, "type", static_cast<int>(type));
+    ospSet2i(ospTexture, "size", size.x, size.y);
+    const auto texelBytes  = ospray::sizeOf(type);
+    const auto totalTexels = size.x * size.y;
+    const auto totalBytes  = totalTexels * texelBytes;
```
tribal-tec: add a function to Texture2D to get sizeInBytes()? We should even reuse that for the scene size statistics, but your choice.

engines/ospray/ispc/geometry/ExtendedSpheres.h
```diff
@@ -28,7 +28,7 @@ namespace ospray
 {
 struct ExtendedSpheres : public ospray::Geometry
 {
-    std::string toString() const final { return "hbp::ExtendedSpheres"; }
+    std::string toString() const final { return "ospray::ExtendedSpheres"; }
```
tribal-tec: more like brayns/bbp namespace?

engines/ospray/OSPRayMaterial.cpp
hernando: I can add the function, but I'm unsure about the future of this code anyway. As far as I know the 2D texture was only used for simulation and we are going to change that anyway, aren't we?

CMakeLists.txt
hernando: I'll remove it, but experience says that it won't work.

karjonas:
LGTM. Did you also try optix? You need to enable it with BRAYNS_OPTIX_ENABLED and OptiX_INSTALL_DIR flags.

Dockerfile
hernando: Because v1.7.0 won't work and they haven't created the v1.7.1 tag yet.

CMakeLists.txt
hernando: Indeed

PR #593, opened 2018-10-18
Do not compile and add empty loaders

karjonas:
If the dependencies for a loader is not met we do not compile and add it to the loader registry.

Move all archive extraction code to archive loader.


brayns/Brayns.cpp
```diff
@@ -58,16 +56,25 @@
 #include <plugins/DeflectPlugin/DeflectPlugin.h>
 #endif
 
-#ifdef BRAYNS_USE_BRION
+#if BRAYNS_ASSIMP_ENABLED
+#include <brayns/io/MeshLoader.h>
+#endif
+
+#if BRAYNS_USE_BRION
+#if BRAYNS_ASSIMP_ENABLED
```
tribal-tec: hmm, ideally the circuit loading would work w/o assimp as well, if you consider only using simple geometries and/or sdf. I guess that's not easily doable right now because of the tight coupling inside the circuit loader.

karjonas: I think I can do it with using the defines actually.

CMakeLists.txt
```diff
@@ -56,6 +56,16 @@ if(libuv_FOUND)
...
 endif()
 
+if(LibArchive_FOUND)
+  option(BRAYNS_LIBARCHIVE_ENABLED "Activate libarchive for archive loading" ON)
+else()
+  unset(BRAYNS_LIBARCHIVE_ENABLED)
+endif()
+
+if(BRAYNS_LIBARCHIVE_ENABLED)
+  add_definitions(-DBRAYNS_LIBARCHIVE_ENABLED=1)
```
tribal-tec: use `common_find_package_disable(assimp)` to unset the USE_ASSIMP var

karjonas:
retest this please

karjonas:
retest this please

PR #594, opened 2018-10-18
Transfer function per model

tribal-tec:
- new RPCs for get- and set-model-transfer-function
- simulation handler per model
- create volumes from model, not from scene
- also remove SpikeReport after NESTLoader removal
- remove dead volume and transfer function code from Optix
- remove simulation histogram and HTTP endpoint
- remove loading of transfer function from file


engines/ospray/OSPRayModel.h
```diff
@@ -22,7 +22,7 @@
 #define OSPRAYGEOMETRYGROUP_H
```
karjonas: Unrelated, but this define does not match the file name.

engines/ospray/OSPRayModel.cpp
```diff
@@ -43,22 +48,52 @@ OSPData allocateVectorData(const std::vector<VecT>& vec,
     return ospNewData(totBytes / ospray::sizeOf(ospType), ospType, vec.data(),
                       memoryManagementFlags);
 }
+
+double interpolatedOpacity(const Vector2ds& controlPoints, const double x)
```
karjonas:
```suggestion
double interpolatedOpacity(const Vector2ds& controlPoints, const double x) const
```

brayns/parameters/AnimationParameters.h
```diff
@@ -78,6 +101,8 @@ class AnimationParameters : public AbstractParameters
     double _dt{0};
     std::string _unit;
 
+    std::list<IsReadyCallback> _isReadyCallbacks;
```
karjonas:
```suggestion
    IsReadyCallbacks _isReadyCallbacks;
```

brayns/parameters/AnimationParameters.h
```diff
@@ -35,37 +36,59 @@ class AnimationParameters : public AbstractParameters
...
     }
     uint32_t getEnd() const { return _end; }
-    void reset()
+    /** The dt of a simulation. */
+    void setDt(const double dt) { _updateValue(_dt, dt); }
+    double getDt() const { return _dt; }
+    /** The time unit of a simulation. */
+    void setUnit(const std::string& unit) { _updateValue(_unit, unit); }
+    using IsReadyCallback = std::function<bool()>;
+    using IsReadyCallbacks = std::list<IsReadyCallback>;
```
karjonas:
```suggestion
    using IsReadyCallbacks = std::list<IsReadyCallback>;  // NOTE: Needs to be list for iterator validity
```

engines/ospray/OSPRayModel.cpp
tribal-tec: free function, const doesn't apply

brayns/parameters/AnimationParameters.h
tribal-tec: Good catch

brayns/parameters/AnimationParameters.h
tribal-tec: Isn't it obvious? why else would one use a list :)

brayns/Brayns.cpp
```diff
@@ -633,26 +608,18 @@ struct Brayns::Impl : public PluginAPI
...
-        }
-
         auto& animParams = _parametersManager.getAnimationParameters();
+        if (!animParams.canUpdateFrame())
+            return;
         const auto animationFrame = animParams.getFrame();
         animParams.setFrame(animationFrame + 1);
     }
 
     void _decreaseAnimationFrame()
```
hernando: Remove and use a -1 as the function parameter for the call suggested above.

brayns/Brayns.cpp
```diff
@@ -633,26 +608,18 @@ struct Brayns::Impl : public PluginAPI
 
     void _increaseAnimationFrame()
```
hernando: Move to AnimationParams::jumpFrames(int frames)

brayns/common/loader/LoaderRegistry.h
```diff
@@ -73,6 +73,6 @@ class LoaderRegistry
               LoaderProgress cb);
 
 private:
-    std::vector<std::unique_ptr<Loader>> _loaders;
+    std::deque<std::unique_ptr<Loader>> _loaders;
```
hernando: Premature optimization?

brayns/common/scene/Model.h
```diff
@@ -199,7 +202,7 @@ class Model
     BRAYNS_API bool empty() const;
 
     /** @return true if the geometry Model is dirty, false otherwise */
-    BRAYNS_API bool dirty() const;
+    BRAYNS_API bool isGeometryDirty() const;
```
hernando: If there's no `isTransferFunctionDirty` you can also call this one `isDirty`

brayns/common/scene/Model.h
```diff
@@ -328,6 +331,24 @@ class Model
...
+        const DataType type) const = 0;
+
+    /**
+     * Create a volume with the given dimensions, voxel spacing and data type
+     * where the voxels are copied via setBrick() into an optimized internal
+     * storage.
+     */
+    BRAYNS_API virtual BrickedVolumePtr createBrickedVolume(
+        const Vector3ui& dimensions, const Vector3f& spacing,
+        const DataType type) const = 0;
```
hernando: SharedDataVolume and BrickedVolume look like OSPRay specific details leaking through the abstractions, or am I wrong?

plugins/RocketsPlugin/jsonSerialization.h
```diff
@@ -257,35 +276,33 @@ inline void init(brayns::ImageGenerator::ImageBase64* i, ObjectHandler* h)
...
-                    reinterpret_cast<std::vector<std::array<float, 4>>*>(
-                        &t->getDiffuseColors()),
-                    Flags::Optional);
-    h->add_property("emission",
-                    reinterpret_cast<std::vector<std::array<float, 3>>*>(
-                        &t->getEmissionIntensities()),
-                    Flags::Optional);
-    h->add_property("contribution", &t->getContributions(), Flags::Optional);
+    h->add_property("range", toArray(t->_valuesRange), Flags::Optional);
+    h->add_property("data", toArray(t->_controlPoints), Flags::Optional);
```
hernando:
```suggestion
    h->add_property("opacityCurve", toArray(t->_controlPoints), Flags::Optional);
```

plugins/RocketsPlugin/RocketsPlugin.cpp
```diff
@@ -459,6 +461,44 @@ class RocketsPlugin::Impl
...
+
+            const auto modelID = document["id"].GetInt();
+            auto model = _engine->getScene().getModel(modelID);
+            if (!model)
+            {
+                return Response{
+                    Response::Error{"Model not found", MODEL_NOT_FOUND}};
+            }
+
+            Document propertyDoc;
```
hernando: A comment about what are you trying to do here makes reading code easier (and avoids guesses)

engines/ospray/ispc/render/AdvancedSimulationRenderer.ispc
```diff
@@ -41,7 +41,6 @@ struct AdvancedSimulationRenderer
     float detectionDistance;
 
     // Volumes
-    uint32 volumeSamplesPerRay;
```
hernando: Where did this go?

engines/ospray/OSPRayModel.h
```diff
@@ -31,28 +30,47 @@ namespace brayns
...
     OSPModel getBoundingBoxModel() const { return _boundingBoxModel; }
     OSPModel getSimulationModel() const { return _simulationModel; }
     MaterialPtr createMaterial(const size_t materialId,
                                const std::string& name) final;
+    SharedDataVolumePtr createSharedDataVolume(const Vector3ui& dimensions,
+                                               const Vector3f& spacing,
+                                               const DataType type) const final;
+    BrickedVolumePtr createBrickedVolume(const Vector3ui& dimensions,
+                                         const Vector3f& spacing,
+                                         const DataType type) const final;
```
hernando: Why did you move this functions from the scene here?

engines/ospray/OSPRayModel.cpp
hernando: This is assuming control points are sorted by x, is that guaranteed by the caller? I haven't seen anything sorting the values received from json for the moment.

brayns/common/loader/LoaderRegistry.h
tribal-tec: no, need `push_front()`

engines/ospray/ispc/render/AdvancedSimulationRenderer.ispc
tribal-tec: was unused

engines/ospray/OSPRayModel.h
tribal-tec: because the volumes need a reference to a transfer function, and this is per model now. Leaving this function here would need a modelID as well or the transfer function itself that has to be used for the new volume.

brayns/common/scene/Model.h
tribal-tec: Yes and no. OSPRay is the one who provides those two volume types (in fact, they support even more). But those two types are still abstract in the sense if you have a simple raw volume (SharedData) or bricked volume format like UVF or BBIC.

brayns/parameters/AnimationParameters.h
karjonas: Some oblivious coder might change it not realizing the consequences ;)

brayns/Brayns.cpp
tribal-tec: done

brayns/Brayns.cpp
tribal-tec: done

plugins/RocketsPlugin/RocketsPlugin.cpp
tribal-tec: done

engines/ospray/OSPRayModel.cpp
tribal-tec: added the sorting, good catch!

tribal-tec:
retest this please

brayns/common/transferFunction/TransferFunction.cpp
```diff
@@ -24,20 +24,34 @@
...
+    if (this == &rhs)
+        return true;
+    return name == rhs.name && colors == rhs.colors;
+}
+
+void ColorMap::clear()
+{
+    colors.clear();
+    colors.push_back({0, 0, 0});
+    colors.push_back({1, 1, 1});
```
hernando: color = {{0, 0, 0}, {1, 1, 1}}

brayns/common/transferFunction/TransferFunction.cpp
```diff
@@ -24,20 +24,34 @@
...
 
 void TransferFunction::clear()
 {
-    _diffuseColors.clear();
-    _emissionIntensities.clear();
-    _contributions.clear();
-    _valuesRange = Vector2f();
+    _controlPoints.clear();
+    _controlPoints.push_back({0, 0});
+    _controlPoints.push_back({1, 1});
```
hernando: _controlPoints = {{0, 0}, {1, 1}}

brayns/io/CircuitLoader.cpp
```diff
@@ -192,11 +200,19 @@ class CircuitLoader::Impl
...
-                    scene.setSimulationHandler(nullptr);
+            if (simulationHandler)
+            {
+                auto& ap = _animationParameters;
+                auto handle = ap.addIsReadyCallback([simulationHandler] {
+                    return simulationHandler->isReady();
                 });
+                ap.setDt(simulationHandler->getDt());
+                ap.setUnit(simulationHandler->getUnit());
+                ap.setEnd(simulationHandler->getNbFrames());
```
hernando: Multiple simulation handlers will not work unless they have the same dt and nbframes. Some consistency needs to be ensured. Also this code may be called concurrently (even if nonsensical), addIsReadyCallback needs  to be thread safe.

Moving this logic out of the constructor may also help.

Or prevent adding a second handler to the animation parameters.

brayns/io/CircuitLoader.cpp
```diff
@@ -38,9 +39,11 @@ namespace brayns
 class CircuitLoader::Impl
 {
 public:
-    Impl(const ApplicationParameters& applicationParameters,
+    Impl(AnimationParameters& animationParameters,
```
hernando: Consider moving the animationParameters out of here because this is not related to model loading, but attaching a model to the scene.

engines/ospray/OSPRayModel.h
```diff
@@ -31,28 +30,48 @@ namespace brayns
...
     void _commitCones(const size_t materialId);
     void _commitMeshes(const size_t materialId);
     void _commitStreamlines(const size_t materialId);
     void _commitSDFGeometries();
+    bool _commitTransferFunction();
+    bool _commitSimulationData();
+
+    AnimationParameters& _animationParameters;
+    VolumeParameters& _volumeParameters;
+    bool _setIsReadyCallback{false};
```
hernando:
```suggestion
    bool _setIsReadyCallback{false};  // Whether this model has set the AnimationParameters "is ready" callback
```

tribal-tec: done

PR #596, opened 2018-10-22
Improve loader output

karjonas:
* Print name of invalid files instead of 'Unsupported Type'

* Print basic progress output

* Check that all files are valid before loading any


karjonas:
retest this please

PR #597, opened 2018-10-22
Add a target state to the camera. Bug fixes.

chevtche:


brayns/common/camera/FlyingModeManipulator.cpp
```diff
@@ -63,14 +63,18 @@ void FlyingModeManipulator::dragLeft(const Vector2i& to, const Vector2i& from)
 
 void FlyingModeManipulator::dragRight(const Vector2i& to, const Vector2i& from)
 {
-    const float distance = -(to.y() - from.y()) * getMotionSpeed();
+    const float distance = -(to.y() - from.y()) *
+                           DEFAULT_MOUSE_MOTION_SPEED_MULTIPLIER *
```
hernando: This looks like getMotionSpeed and getRotationSpeed should return smaller values. I would do that instead of adding this multiplication by DEFAULT_MOUSE_... everywhere.

I have to test myself how your manipulator works now before judging if this is an improvement.

chevtche: No you can't. If you do so keyboard mouvements are to slow. Believe me, it's an improvement :)

PR #598, opened 2018-10-23
Support subsampling again after OSPRay 1.7 upgrade

tribal-tec:
This implies that the subsampled framebuffer which is used during interaction (and a
negative samples-per-pixel value), is now smaller than the window/viewport size.

Hence, clients are supposed to scale the image/framebuffer. This is done for the
braynsViewer using a texture.

### DIFF
--- a/apps/ui/BaseWindow.cpp
+++ b/apps/ui/BaseWindow.cpp
@@ -136,9 +136,7 @@ BaseWindow::BaseWindow(Brayns& brayns, const FrameBufferMode frameBufferMode)
     BRAYNS_INFO << "Motion speed :" << motionSpeed << std::endl;
 }
 
-BaseWindow::~BaseWindow()
-{
-}
+BaseWindow::~BaseWindow() {}
 
 void BaseWindow::mouseButton(const int button, const bool released,
                              const Vector2i& pos)
@@ -300,8 +298,30 @@ void BaseWindow::display()
 
         if (buffer)
         {
-            glDrawPixels(renderOutput.frameSize.x(), renderOutput.frameSize.y(),
-                         format, type, buffer);
+            glMatrixMode(GL_PROJECTION);
+            glLoadIdentity();
+            glOrtho(0.0f, windowSize.x(), 0.0f, windowSize.y(), -1.0f, 1.0f);
+
+            glMatrixMode(GL_MODELVIEW);
+            glLoadIdentity();
+            glViewport(0, 0, windowSize.x(), windowSize.y());
+
+            glBindTexture(GL_TEXTURE_2D, _fbTexture);
+            glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA, renderOutput.frameSize.x(),
+                         renderOutput.frameSize.y(), 0, GL_RGBA, type, buffer);
+
+            glBegin(GL_QUADS);
+            glTexCoord2f(0.f, 0.f);
+            glVertex3f(0, 0, 0);
+            glTexCoord2f(0.f, 1.f);
+            glVertex3f(0, windowSize.y(), 0);
+            glTexCoord2f(1.f, 1.f);
+            glVertex3f(windowSize.x(), windowSize.y(), 0);
+            glTexCoord2f(1.f, 0.f);
+            glVertex3f(windowSize.x(), 0, 0);
+            glEnd();
+
+            glBindTexture(GL_TEXTURE_2D, 0);
         }
     }
     else
@@ -363,20 +383,6 @@ void BaseWindow::clearPixels()
     glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
 }
 
-void BaseWindow::drawPixels(const int* framebuffer)
-{
-    glDrawPixels(_windowSize.x(), _windowSize.y(), GL_RGBA, GL_UNSIGNED_BYTE,
-                 framebuffer);
-    glutSwapBuffers();
-}
-
-void BaseWindow::drawPixels(const Vector3f* framebuffer)
-{
-    glDrawPixels(_windowSize.x(), _windowSize.y(), GL_RGBA, GL_FLOAT,
-                 framebuffer);
-    glutSwapBuffers();
-}
-
 void BaseWindow::setTitle(const char* title)
 {
     assert(_windowID >= 0);
@@ -406,6 +412,13 @@ void BaseWindow::create(const char* title, const size_t width,
         _screenSpaceProcessor.init(width, height);
 
     _registerKeyboardShortcuts();
+
+    glGenTextures(1, &_fbTexture);
+    glEnable(GL_TEXTURE_2D);
+    glBindTexture(GL_TEXTURE_2D, _fbTexture);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR);
+    glBindTexture(GL_TEXTURE_2D, 0);
 }
 
 void BaseWindow::keypress(const char key, const Vector2f&)
@@ -417,6 +430,7 @@ void BaseWindow::keypress(const char key, const Vector2f&)
         break;
     case 27:
     case 'Q':
+        _onExit();
 #ifdef __APPLE__
         exit(0);
 #else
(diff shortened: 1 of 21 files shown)


tribal-tec:
@rolandjitsu Needs upscaling of the JPEG image to the `viewport` size

apps/ui/BaseWindow.cpp
```diff
@@ -300,8 +298,30 @@ void BaseWindow::display()
...
+            glVertex3f(0, 0, 0);
+            glTexCoord2f(0.f, 1.f);
+            glVertex3f(0, windowSize.y(), 0);
+            glTexCoord2f(1.f, 1.f);
+            glVertex3f(windowSize.x(), windowSize.y(), 0);
+            glTexCoord2f(1.f, 0.f);
+            glVertex3f(windowSize.x(), 0, 0);
+            glEnd();
+
+            glBindTexture(GL_TEXTURE_2D, 0);
```
hernando: Maybe it's asking too much, but can you avoid using non GL3 core functions? You'll need shaders, but they are quite trivial.

apps/ui/BaseWindow.cpp
```diff
@@ -406,6 +412,13 @@ void BaseWindow::create(const char* title, const size_t width,
         _screenSpaceProcessor.init(width, height);
 
     _registerKeyboardShortcuts();
+
+    glGenTextures(1, &_fbTexture);
+    glEnable(GL_TEXTURE_2D);
+    glBindTexture(GL_TEXTURE_2D, _fbTexture);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR);
+    glBindTexture(GL_TEXTURE_2D, 0);
```
hernando: No glDeleteTextures?

apps/ui/BaseWindow.cpp
```diff
@@ -406,6 +412,13 @@ void BaseWindow::create(const char* title, const size_t width,
         _screenSpaceProcessor.init(width, height);
 
     _registerKeyboardShortcuts();
+
+    glGenTextures(1, &_fbTexture);
+    glEnable(GL_TEXTURE_2D);
+    glBindTexture(GL_TEXTURE_2D, _fbTexture);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR);
```
hernando: MIN filter not needed?

engines/ospray/OSPRayFrameBuffer.cpp
```diff
@@ -159,14 +165,42 @@ void OSPRayFrameBuffer::_unmapUnsafe()
...
+{
+    return _useSubsampling() ? _subsamplingFrameBuffer : _frameBuffer;
+}
+
+Vector2ui OSPRayFrameBuffer::_subsamplingSize() const
+{
+    return _frameSize / _subsamplingFactor;
+}
+
+OSPFrameBufferFormat OSPRayFrameBuffer::_ospFrameBufferFormat() const
```
hernando: Maybe free function with 1 parameter?

engines/ospray/OSPRayFrameBuffer.cpp
```diff
@@ -90,42 +89,49 @@ void OSPRayFrameBuffer::_recreate()
...
+        ospCommit(_subsamplingFrameBuffer);
+    }
+}
+
 void OSPRayFrameBuffer::clear()
 {
     FrameBuffer::clear();
     size_t attributes = OSP_FB_COLOR | OSP_FB_DEPTH;
+    if (_subsamplingFrameBuffer)
+        ospFrameBufferClear(_subsamplingFrameBuffer, attributes);
```
hernando: Once created the _subsamplingFrameBuffer is never destroyed, is it?
Do you need to clear both buffers? Why not just _currentFB?

engines/ospray/OSPRayFrameBuffer.cpp
```diff
@@ -90,42 +89,49 @@ void OSPRayFrameBuffer::_recreate()
...
     ospCommit(_frameBuffer);
+
+    _recreateSubsamplingBuffer();
+
     clear();
 }
 
+void OSPRayFrameBuffer::_recreateSubsamplingBuffer()
+{
+    ospRelease(_subsamplingFrameBuffer);
```
hernando: Just for confirmation, ospRelease(nullptr) is safe, isn't it?

engines/ospray/OSPRayFrameBuffer.cpp
```diff
@@ -54,6 +51,8 @@ void OSPRayFrameBuffer::enableDeflectPixelOp()
     {
         ospCommit(_pixelOp);
         ospSetPixelOp(_frameBuffer, _pixelOp);
+        if (_subsamplingFrameBuffer)
+            ospSetPixelOp(_subsamplingFrameBuffer, _pixelOp);
```
hernando: I don't know the logic of the pixelop, but it seems to me that Deflect may not work properly after subsampling has been enabled.

karjonas:
Is it feasible to return a framebuffer of the correct size so that the client does not need to do any scaling?

hernando:
I guess that's doable using an ispc kernel. However for network streaming I think it's preferable to ask the client to do the upscaling (which can also be hardware accelerated in the display side).

Even without network communication, I would bet that doing the upscaling during display brings better performance as any GPU has better memory bandwidth for such an operation than a CPU and you still have to upload the pixels to the GPU, so the smaller the buffer, the better.

apps/ui/BaseWindow.cpp
tribal-tec: does this really help? I bluntly took what the ospray example viewer is doing, so no efforts spend so far. GL is dead anyways, so using legacy GL is not too much of a difference to GL3.

apps/ui/BaseWindow.cpp
tribal-tec: that's probably missing, thanks

engines/ospray/OSPRayFrameBuffer.cpp
tribal-tec: `clear()` shall clear everything, also the accumulation buffer in the 'normal' framebuffer. `_currentFB()` at the time when `clear()` is called will only clear the subsampling buffer.

engines/ospray/OSPRayFrameBuffer.cpp
tribal-tec: yes, it is. hence I removed some of the checks I found.

engines/ospray/OSPRayFrameBuffer.cpp
tribal-tec: The pixelop code probably might need an adaption. I have to check that, good point. Not sure if I can test it though with my current setup though.

tribal-tec:
I also agree that a server-side upscaling is definetly possible, but actually compressing and sending a smaller image (which in fact it is and it always was even before) makes more sense, plus as Juan mentioned, GPU texture scaling has no performance impact. The subsampling is an optional feature anyways, so clients can easily opt-out if they don't want to deal with that.

apps/ui/BaseWindow.cpp
tribal-tec: MAG is not needed :)

hernando: No, no, it's MIN, I'm sure. You are mapping the texture to a bigger area.

tribal-tec:
Updated

tribal-tec:
retest this please

engines/ospray/OSPRayFrameBuffer.cpp
tribal-tec: The 'normal' deflect streaming works, but the pixelop image is too small.

hernando:
Daniel mentioned that Deflect streaming was not working but he didn't know what was the problem. I'd prefer if we merge this only when that's fixed. The question is whether we wait for him or somebody gives it a try.

rolandjitsu:
We can wait 😄

tribal-tec:
I added the DeflectPixelOp with an according fix for arbitrary (non TILE_SIZE) resolutions, which was the breaking part of this PR. This was an issue as the subsampling buffer can have any resolution in fact.

tribal-tec:
So feel free to review (again) and finally approve :smile:

engines/ospray/OSPRayEngine.cpp
```diff
@@ -198,11 +198,27 @@ void OSPRayEngine::preRender()
...
+        }
+        return result;
+    }
+
+    Vector2f result = size;
+    if (isSideBySideStereo)
+    {
+        if (size.x() % (TILE_SIZE * 2) != 0)
+            result.x() = size.x() - size.x() % (TILE_SIZE * 2);
+    }
```
hernando: Use the simpler code from the previous commit

PR #599, opened 2018-10-23
Add 'default-bvh-flag' command line option

karjonas:


brayns/common/types.h
```diff
@@ -373,6 +373,14 @@ struct RpcParameterDescription : RpcDescription
     std::string paramName;
     std::string paramDescription;
 };
+
+enum class BVHType
+{
+    dynamic,
```
tribal-tec: that's very ospray centric. what is 'none' even? Shouldn't it be a bitmask also? dynamic + compact sounds reasonable to me. If ospray can't do that, we shouldn't limit it here. 'Static' is missing also to me.

engines/ospray/OSPRayModel.cpp
```diff
@@ -510,6 +510,29 @@ bool OSPRayModel::_commitTransferFunction()
     return true;
 }
 
+void OSPRayModel::_commitBVHType()
```
tribal-tec: `ospCommit(_model)` is missing then, otherwise the function should be renamed.

brayns/common/types.h
hernando: Since optimizations for a particular metric typically go against others, I think it's fair to have a single label to provide a hint about the expected usage.

karjonas:
How do I query the memory usage of the BVH?

brayns/common/utils/Utils.h
```diff
@@ -34,4 +36,18 @@ inline auto lowerCase(std::string str)
...
+
+template <typename T>
+inline T enumSet(const T eA, const T eB)
+{
+    using U = typename std::underlying_type<T>::type;
+    return static_cast<T>(static_cast<U>(eA) | static_cast<U>(eB));
+}
+
+template <typename T>
+inline bool enumHas(const T eA, const T eB)
```
hernando: template<typename T>
    bool isBitSet(typename std::underlying_type<T>::type mask, const T bit)

brayns/common/utils/Utils.h
```diff
@@ -34,4 +36,18 @@ inline auto lowerCase(std::string str)
     std::transform(str.begin(), str.end(), str.begin(), ::tolower);
     return str;
 }
+
+template <typename T>
+inline T enumSet(const T eA, const T eB)
```
hernando:
```suggestion
template <typename T>
inline typename std::underlying_type<T>::type setBit(const std::underlying_type<T>::type mask, const T bit)
{
     return mask | static_cast<typename std::underlying_type<T>::type>(bit);
}
```

You should return the underlying type, because semantically the output value is not the enum anymore.
But even worse, imagine you have this code

    enum class A : int { a = 1, b = 2 };
    int f(const A a)
    {
        switch(a)
        {
            case A::a: return int(a);
            case A::b: return int(a);
        }
    }

That code if perfect, but if you call f with something A(17), the function return value will be undefined.

brayns/parameters/GeometryParameters.cpp
```diff
@@ -194,7 +204,11 @@ GeometryParameters::GeometryParameters()
         //
         (PARAM_CIRCUIT_MESH_TRANSFORMATION.c_str(),
          po::bool_switch()->default_value(false),
-         "Enable mesh transformation according to circuit information.");
+         "Enable mesh transformation according to circuit information.")
+        //
+        (PARAM_DEFAULT_BVH_FLAG.c_str(),
+         po::value<std::vector<std::string>>()->multitoken(),
+         "Default bvh type [static|dynamic|compact|robust]");
```
hernando: It's not clear that you can have multiple flags and what is the token separator.

hernando: BVH

brayns/parameters/GeometryParameters.h
```diff
@@ -234,6 +234,7 @@ class GeometryParameters : public AbstractParameters
         return _morphologyUseSDFGeometries;
     }
 
+    BVHType getDefaultBVHType() const { return _defaultBVHType; }
```
hernando: int getDefaultBVHModes or getDefaultBVHTypeMask

engines/ospray/OSPRayModel.cpp
```diff
@@ -510,6 +512,17 @@ bool OSPRayModel::_commitTransferFunction()
...
 
+void OSPRayModel::_setBVHType()
+{
+    const bool dynamicMode = enumHas(_bvhType, BVHType::dynamic);
+    const bool compactMode = enumHas(_bvhType, BVHType::compact);
+    const bool robustMode = enumHas(_bvhType, BVHType::robust);
+
+    ospSet1i(_model, "dynamicMode", dynamicMode ? 1 : 0);
+    ospSet1i(_model, "compactMode", compactMode ? 1 : 0);
+    ospSet1i(_model, "robustMode", robustMode ? 1 : 0);
```
hernando:
```suggestion
    ospSet1i(_model, "robustMode", isBitSet(_bvhType, BVHTtype::dynamic));
```

brayns/common/scene/Model.h
```diff
@@ -97,9 +97,12 @@ class ModelParams : public ModelInstance
     const std::string& getName() const { return _name; }
     void setPath(const std::string& path) { _updateValue(_path, path); }
     const std::string& getPath() const { return _path; }
+    void setBVHType(BVHType bvhType) { _updateValue(_bvhType, bvhType); }
+    BVHType getBVHType() const { return _bvhType; }
```
hernando: It would be better if this parameter goes into the ModelDescriptor property map because then you can set it from the client. I'm not totally sure about the jsonrpc part, maybe @rolandjitsu can comment on this.

rolandjitsu: If you add it to the property map it would be rendered without any effort on my side (because it's generated from schema).

But then you have to make sure that when the bvh property on the property map is set, you do your stuff (set the bvh type).

Also, I'm not sure how enums are handled currently for the property map props (if they are).

hernando: In fact it's not an enum, it's a bit mask. From the command line I think it's set as a space separated list of strings. I'd guess you don't have any code for rendering such a thing.

rolandjitsu: Nope, we don't handle bit masks.

For multiple selection we'd need some additional info from schema (instead of `{type: 'string', enum: ['bla', ...]}` has to be `{type: 'array', items: {type: 'string', enum: ['bla', ...]}`).

And then I'd need to add support for multiple selection.

If this is single selection, then we just need string enums instead of bitmask.

hernando: @tribal-tec requested it to be a multiple selection, but not all combinations are allowed... Another option is splitting this into multiple properties:
- BVH update hint: static|dynamic
- compact BVH: bool
- robust BVH: bool
What do you think @karjonas ?

hernando: This should be easy to render and we only need to extend the schema to support single selection for a set of values.

hernando: As discussed offline, let's keep it as it is for the moment and open this can of worms later.

brayns/common/scene/Model.h
```diff
@@ -400,7 +402,11 @@ class Model
     void resetVolumesDirty() { _volumesDirty = false; }
     /** @internal */
     void updateSizeInBytes();
-
+    void setBVHFlags(std::set<BVHFlag> bvhFlags)
+    {
+        _bvhFlags = std::move(bvhFlags);
+    }
+    std::set<BVHFlag> getBVHFlags() const { return _bvhFlags; }
```
hernando: `const std::set<BVHFlag>&`

brayns/parameters/GeometryParameters.cpp
```diff
@@ -194,7 +199,12 @@ GeometryParameters::GeometryParameters()
         //
         (PARAM_CIRCUIT_MESH_TRANSFORMATION.c_str(),
          po::bool_switch()->default_value(false),
-         "Enable mesh transformation according to circuit information.");
+         "Enable mesh transformation according to circuit information.")
+        //
+        (PARAM_DEFAULT_BVH_FLAG.c_str(),
+         po::value<std::vector<std::string>>()->multitoken(),
+         "Sets one or more property flags to be used by default when creating "
+         "a bvh [dynamic|compact|robust]");
```
hernando: BVH
Same as before, it's not clear that this is a list? Or does boost program_options take care of indicating it?

brayns/parameters/GeometryParameters.cpp
```diff
@@ -332,6 +342,20 @@ void GeometryParameters::parse(const po::variables_map& vm)
...
+    {
+        const auto& bvhs =
+            vm[PARAM_DEFAULT_BVH_FLAG].as<std::vector<std::string>>();
+        for (const auto& bvh : bvhs)
+        {
+            const auto kv = BVH_TYPES.find(bvh);
+            if (kv != BVH_TYPES.end())
+                _defaultBVHFlags.insert(kv->second);
+            else
+                throw std::runtime_error("Invalid bvh type '" + bvh + "'.");
```
hernando: type -> flag

engines/ospray/OSPRayModel.cpp
```diff
@@ -510,6 +510,17 @@ bool OSPRayModel::_commitTransferFunction()
     return true;
 }
 
+void OSPRayModel::_setBVHFlags()
+{
+    const bool dynamicMode = _bvhFlags.count(BVHFlag::dynamic);
+    const bool compactMode = _bvhFlags.count(BVHFlag::compact);
+    const bool robustMode = _bvhFlags.count(BVHFlag::robust);
+
+    ospSet1i(_model, "dynamicMode", dynamicMode ? 1 : 0);
```
hernando:
```suggestion
    ospSet1i(_model, "dynamicMode", _bvhFlags.count(BVHFlag::dynamic));
```

tests/brayns.cpp
```diff
@@ -120,3 +121,21 @@ BOOST_AUTO_TEST_CASE(defaults)
...
     BOOST_CHECK(geomParams.getMemoryMode() == brayns::MemoryMode::shared);
 }
+
+BOOST_AUTO_TEST_CASE(bvh_type)
+{
+    auto& testSuite = boost::unit_test::framework::master_test_suite();
+    const char* app = testSuite.argv[0];
+    const char* argv[] = {
+        app,       "demo", "--default-bvh-flag", "robust", "--default-bvh-flag",
+        "compact",
```
hernando: I would have never guessed this is how you provide multiple flags through the command line...

brayns/parameters/GeometryParameters.cpp
karjonas: I don't know how to explain it without a wall of text.

hernando: "Set a default flag to apply to BVH creation, one of  [dynamic|compact|robust], may appear multiple times."

I'm not sure whether --default-bvh-flag "dynamic robust" is also accepted though.

karjonas: It is.

hernando: Then, the message is not totally accurate, but at least it doesn't miss the possibility of setting multiple flags.

PR #600, opened 2018-11-02
Geometry and materials

hernando:
This PR includes the removal of spheres and cylinders from Brayns to use OSPRay geometries instead. The remaining classes has been stripped off the Extended prefix.

The other big change is the refactoring of material handling to not use texture coordinates for simulation mapping and use ad hoc offsets that make part of the geometry data array.


engines/ospray/ispc/render/SimulationMaterial.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include <ospray/SDK/geometry/Spheres.h>
+
+#include <map>
+#include <typeindex>
+#include <typeinfo>
+
+extern "C" {
+int SimulationMaterial_getBytesPerPrimitive(const void* geometry)
+{
+    const ospray::Geometry* base =
```
karjonas: How often is this called? I am concerned about the performance of doing several dynamic casts like this.

hernando: It's called every shaded hit. The dynamic cast is not ideal but I started implementing another alternative using std::type_index that I though it would be faster, but it won't as it requires hashing of strings, something which is for sure slower than a dynamic_cast. I haven't seen any noticeable performance difference in the experiments I've done with a debug build.

In any case, I'd go for correctness first and optimizations later in a separate PR.

hernando: A potentially faster solution would be to keep track of all geometry instances in a std::unordered_map<void*, size_t> that stores what this function should return. Hashing a pointer and doing a lookup on this table seems faster than RTTI. However, the code to keep track of the instances will be a lot dirtier.

brayns/common/scene/Scene.h
```diff
@@ -229,6 +229,9 @@ class Scene : public BaseObject
                                  const size_t materialID,
                                  const ModelParams& params, LoaderProgress cb);
 
+    template<typename T>
+    inline void visitModels(const T& functor);
```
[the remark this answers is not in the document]
hernando: I guess it's not performance critical so I'd do it.

hernando:
The tests don't pass due to an issue in OSPRay (ospray/ospray#279). With the proposed change in that issue there's only one test that doesn't pass, but the difference is small and can be reduced scaling down the epsilon. It doesn't look like the issue is going to be solved quickly, so I would merge this change once it has a positive review.

tests/PDiffHelpers.h
```diff
@@ -61,25 +58,15 @@ inline bool compareTestImage(const std::string& filename,
...
-#ifdef GENERATE_TESTDATA
-    {
-        auto image = brayns::freeimage::getImageFromFrameBuffer(fb);
-        FreeImage_Save(FreeImage_GetFIFFromFilename(filename.c_str()),
-                       image.get(), fullPath.c_str());
-    }
-#endif
+    static bool saveImages = getenv("BRAYNS_SAVE_TEST_IMAGES");
 
-#ifdef WRITE_TEMP_IMAGES
```
karjonas: Is there a reason for disabling the temp images, I find them helpful at times?

engines/ospray/ispc/render/SimulationMaterial.cpp
karjonas: Is it possible to store an enum in the user data of the geometry?

tests/PDiffHelpers.h
karjonas: Okay, makes sense.

engines/ospray/ispc/render/SimulationMaterial.cpp
hernando: Hahaha, is that a joke?

There's no such a think as Geometry user data, I wished it existed, otherwise I wouldn't have spent so much time with ispc generating stupid code for something it shouldn't accept as correct code in the first place.

karjonas: Sadly, it is not a joke. I am confused by this since you do get the geometry data pointer like this:

```
   // The data pointer in all "derived" geometries is just after data members
    // of the base Geometry struct.
    const uniform uint8* data =  *((const uniform uint8**)&geometry[1]);
```

So could you not put the enum first then?

hernando: No, because the Geometry struct is created by ISPC code in OSPRay. There's no way I can change or tinker with that. My code is already type unsafe as it's making an assumption about the location of this data pointer, so don't think I'm happy with it, but I didn't find a better way without modifying OSPRay or going back to have our own spheres and cylinders.

karjonas: Okay, we leave it like this then.

engines/ospray/OSPRayModel.cpp
```diff
@@ -592,6 +537,46 @@ bool OSPRayModel::commitTransferFunction()
...
+        {
+            auto matIt = _materials.begin();
+            auto geomIt = map.begin();
+            while (matIt != _materials.end() && geomIt != map.end())
+            {
+                while (matIt->first < geomIt->first)
+                    ++matIt;
+                if (matIt->first != geomIt->first)
+                {
+                    // This shoulnd't happen
```
tribal-tec: typo

brayns/PluginManager.cpp
```diff
@@ -1,6 +1,6 @@
 /* Copyright (c) 2015-2018, EPFL/Blue Brain Project
  * All rights reserved. Do not distribute without permission.
- * Responsible Author: Juan Hernando <cyrille.favreau@epfl.ch>
+ * Responsible Author: Juan Hernando <juan.hernandoepfl.ch>
```
tribal-tec:
```suggestion
 * Responsible Author: Juan Hernando <juan.hernando@epfl.ch>
```

PR #601, opened 2018-11-05
Renamings in plugin directories and targets

hernando:
Removed the Plugin suffix from plugin subdirectories,
Consistent naming for all plugins libraries: braynsFoo instead of braynsFooPlugin or FooPlugin.


hernando:
@ppodhajski, I've renamed libVRPNPlugin to libbraynsVRPN. Let me know if that can cause problems in the deployment.

PR #602, opened 2018-11-07
Do not depend on OSPRay for loading dynamic libraries

hernando:
Removed ExtensionPluginFactory as it was a mere container, not
a factory.


hernando:
I can anticipate that CI is not passing, I didn't test it correctly before submission.

hernando:
Bug fixed, you can approve again.

karjonas:
If we add that const to argv we need to recompile all plugins, right?

hernando:
Not really, the public API hasn't really changed. The function brayns_create_plugin has C linkage so there's no mangling based on argument types. At the same time, the symbol is cast into the expected type no matter what it actually is. You could potentially declare it to be something else and crash at call time as type safety cannot be enforced.

brayns/Brayns.cpp
```diff
@@ -835,6 +832,9 @@ struct Brayns::Impl : public PluginAPI
         app.setSynchronousMode(!app.getSynchronousMode());
     }
 
+    std::vector<DynamicLib> _pluginLibs;
+    std::vector<ExtensionPluginPtr> _extensions;
```
rdumusc: I would move the 2 loadPlugin methods and these members inside the ExtensionPluginFactory (could be renamed to something nicer) instead of removing the class

brayns/common/utils/DynamicLib.cpp
```diff
@@ -0,0 +1,89 @@
+#ifdef _WIN32
```
rdumusc: no copyright in this file?

brayns/Brayns.cpp
hernando: Hmm, maybe at a later stage because this is evolving. I need to see how it works with command line options before reviving it.

PR #603, opened 2018-11-07
Brayns initialization modified

hernando:
Plugin loading now happens in the constructor, as it's needed to register
loaders from plugins before loading occurs (this will be the case of the
upcoming CircuitViewer plugin).

In braynsService, the event loop setup now must be done before creating
the Brayns instance.

### DIFF
--- a/apps/BraynsService/main.cpp
+++ b/apps/BraynsService/main.cpp
@@ -33,8 +33,7 @@ class BraynsService
 {
 public:
     BraynsService(int argc, const char** argv)
-        : _brayns(argc, argv)
-        , _renderingDone{_mainLoop->resource<uvw::AsyncHandle>()}
+        : _renderingDone{_mainLoop->resource<uvw::AsyncHandle>()}
         , _eventRendering{_mainLoop->resource<uvw::IdleHandle>()}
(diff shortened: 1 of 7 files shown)


apps/BraynsService/main.cpp
```diff
@@ -156,7 +156,7 @@ class BraynsService
         _renderLoop->stop();
     }
 
-    brayns::Brayns _brayns;
+    std::shared_ptr<brayns::Brayns> _brayns;
```
rdumusc: why not unique_ptr?

apps/BraynsService/main.cpp
```diff
@@ -47,7 +46,14 @@ class BraynsService
         _setupMainThread();
         _setupRenderThread();
 
-        _brayns.loadPlugins();
+        _brayns.reset(new brayns::Brayns(argc, argv));
```
karjonas:
```suggestion
        _brayns = std::make_unique<brayns::Brayns>(argc, argv);
```

karjonas:
The copyright in DynamicLib.cpp should still be Intel, or did you change the code a lot?

hernando:
Yes, I did

PR #612, opened 2018-11-19
Add support for multiple frameBuffers

tribal-tec:


apps/ui/BaseWindow.cpp
```diff
@@ -243,103 +251,103 @@ void BaseWindow::forceRedraw()
 
 void BaseWindow::display()
 {
-    const Vector2ui windowSize = _brayns.getParametersManager()
-                                     .getApplicationParameters()
-                                     .getWindowSize();
-    if (windowSize != _windowSize)
-        glutReshapeWindow(windowSize.x(), windowSize.y());
+    Vector2ui newWindowSize = _getWindowSize();
```
rdumusc: const auto?

brayns/Brayns.cpp
```diff
@@ -269,6 +271,38 @@ struct Brayns::Impl : public PluginAPI
...
+            _engine->addFrameBuffer(rightBuffer);
+            _frameBuffers.push_back(rightBuffer);
+        }
+        else
+        {
+            auto frameBuffer =
+                _engine->createFrameBuffer("default", frameSize,
+                                           FrameBufferFormat::rgba_i8);
+            _engine->addFrameBuffer(frameBuffer);
+            _frameBuffers.push_back(frameBuffer);
```
rdumusc: 3 times the same code, make a small helper function:
_createFrameBuffer(const std::string& name)

brayns/common/engine/Engine.cpp
```diff
@@ -74,23 +83,48 @@ Renderer& Engine::getRenderer()
...
+{
+    _frameBuffers.push_back(frameBuffer);
+}
+
+void Engine::removeFrameBuffer(FrameBufferPtr frameBuffer)
+{
+    auto i = std::find_if(_frameBuffers.begin(), _frameBuffers.end(),
+                          [&frameBuffer](auto x) { return frameBuffer == x; });
+    if (i != _frameBuffers.end())
+        _frameBuffers.erase(i);
```
rdumusc: doesn't just _frameBuffers.erase(frameBuffer); do the right thing?

brayns/common/renderer/FrameBuffer.h
```diff
@@ -45,6 +44,7 @@ class FrameBuffer : public BaseObject
     virtual void resize(const Vector2ui& frameSize) = 0;
 
     virtual Vector2ui getSize() const { return _frameSize; }
+    const Vector2ui& getFrameSize() const { return _frameSize; }
```
rdumusc: Is it needed to have getFrameSize() in addition to getSize()?

engines/ospray/ispc/camera/PerspectiveCamera.h
```diff
@@ -45,13 +42,7 @@ struct OSPRAY_SDK_INTERFACE PerspectiveCamera : public Camera
...
     bool architectural; // orient image plane to be parallel to 'up' and shift
                         // the lens
-    typedef enum {
-        OSP_STEREO_NONE,
-        OSP_STEREO_LEFT,
-        OSP_STEREO_RIGHT,
-        OSP_STEREO_SIDE_BY_SIDE
-    } StereoMode;
-    StereoMode stereoMode;
+    bool stereo;
```
rdumusc: uninitialized variable (the ones above too!)

brayns/common/engine/Engine.cpp
tribal-tec: It's a `std::vector`, so unfortunately, no.

engines/ospray/ispc/camera/PerspectiveCamera.h
tribal-tec: The initialization for ospray objects to default values is done in `commit()`, where the default values are actually used if the given parameter does not exist. Check the cpp file for that.

brayns/common/renderer/FrameBuffer.h
tribal-tec: Yes, as `getSize()` might be smaller if subsampling is used. Yet the 'real' frame buffer size is needed for the viewer.

plugins/Deflect/DeflectPlugin.cpp
```diff
@@ -273,29 +287,46 @@ class DeflectPlugin::Impl
...
         sizeHints.maxHeight = std::numeric_limits<unsigned int>::max();
-        sizeHints.preferredWidth = size.x();
-        sizeHints.preferredHeight = size.y();
         sizeHints.minWidth = minSize.x();
         sizeHints.minHeight = minSize.y();
 
-        if (_engine.getCamera().isSideBySideStereo())
+        // only send preferred size if we have no multi-channel setup (e.g.
+        // OpenDeck)
+        uint8_t channel = _getChannel(frameBuffers[0]->getName());
```
hernando: const uint8_t channel

plugins/Rockets/SnapshotTask.h
```diff
@@ -56,9 +56,8 @@ class SnapshotFunctor : public TaskFunctor
     SnapshotFunctor(Engine& engine, SnapshotParams&& params,
                     ImageGenerator& imageGenerator)
         : _params(std::move(params))
-        , _frameBuffer(engine.createFrameBuffer(_params.size,
-                                                FrameBufferFormat::rgba_i8,
-                                                true))
+        , _frameBuffer(engine.createFrameBuffer("snapshot", _params.size,
+                                                FrameBufferFormat::rgba_i8))
```
hernando: Check with @favreau how he's generating stereo movies at the moment because it's not possible to get left/right eye images anymore (even less side by side).

plugins/Rockets/RocketsPlugin.cpp
```diff
@@ -977,15 +972,6 @@ class RocketsPlugin::Impl : public ActionInterface
                               "renderer");
     }
 
-    void _handleFrameBuffer()
-    {
-        // don't add framebuffer to websockets for performance
-        using namespace rockets::http;
-        _rocketsServer->handleGET(ENDPOINT_FRAME_BUFFERS,
-                                  _engine.getFrameBuffer());
-        _handleObjectSchema(ENDPOINT_FRAME_BUFFERS, _engine.getFrameBuffer());
```
hernando: This was used in the past to get the depth and color buffers in a Python notebook and do stuff with them for the reprojection prototype. Would it be difficult to change this end point to return a list of buffer names and have an RPC to retrieve a frame buffer by name. Doesn't have to go in this PR.

plugins/Deflect/DeflectPlugin.cpp
```diff
@@ -246,14 +266,8 @@ class DeflectPlugin::Impl
...
                 Vector2ui newSize(event.dx, event.dy);
-                if (_engine.getCamera().isSideBySideStereo())
-                    newSize.x() *= 2;
-
                 if (_params.getResizing())
-                {
-                    _appParams.setWindowSize(
-                        _engine.getSupportedFrameSize(newSize));
-                }
+                    _appParams.setWindowSize(newSize);
```
hernando: I think this is not correct. In braynsViewer (not sure about the service), the window size is the union of the frame buffers and that's what you set in the application parameters, but here the size is supposed to be the size of just a buffer. Does it mean the stereo streaming from braynsService is not supported?

I tried to see what I get in this case myself, but as a matter of fact there seems to be another problem when streaming from braynsViewer as the brayns process window is ping-pong'ing  between two different sizes, even without enabling stereo. On the other hand, braynsService is OK.

brayns/Brayns.cpp
```diff
@@ -120,10 +118,15 @@ struct Brayns::Impl : public PluginAPI
         const auto windowSize =
             _parametersManager.getApplicationParameters().getWindowSize();
 
-        _engine->reshape(windowSize);
+        auto& camera = _engine->getCamera();
+        camera.updateProperty("aspect",
+                              static_cast<double>(windowSize.x()) /
+                                  static_cast<double>(windowSize.y()));
+        for (auto frameBuffer : _frameBuffers)
+            frameBuffer->resize(windowSize);
```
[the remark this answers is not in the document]
tribal-tec: that's what the code does, how does the comment give more/different information?

plugins/Rockets/RocketsPlugin.cpp
hernando: Remove json serialization for FrameBuffer if this is removed.

brayns/Brayns.cpp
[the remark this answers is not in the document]
hernando: For a casual reader, it's not obvious at this point that _frameBuffers is not storing all the frame buffers, but just those that were created by BaseWindow. Spare it if you deemed it unnecessary. I mentioned it because despite it doesn't make sense to store other buffers here, you cannot know it for sure without going to the code where they are assigned, assuming it is just an educated guess, but still a guess.

apps/ui/BaseWindow.cpp
```diff
@@ -153,10 +152,12 @@ void BaseWindow::mouseButton(const int button, const bool released,
 
     if (_currModifiers & GLUT_ACTIVE_SHIFT && released)
     {
+        const auto& windowSize = _brayns.getParametersManager()
+                                     .getApplicationParameters()
+                                     .getWindowSize();
```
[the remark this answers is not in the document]
tribal-tec: In a follow-up commit, the window size shall be removed/renamed to framebuffer size and will be used by plugins/viewer to create buffers in a size they actually need, so no more core-created buffers.

apps/ui/BaseWindow.cpp
```diff
@@ -243,103 +251,103 @@ void BaseWindow::forceRedraw()
...
 
-        ssProcData.depthFormat = GL_LUMINANCE;
-        ssProcData.depthBuffer = renderOutput.depthBuffer.data();
-        ssProcData.depthType = GL_FLOAT;
+            ssProcData.depthFormat = GL_LUMINANCE;
+            ssProcData.depthBuffer = frameBuffer->getDepthBuffer();
+            ssProcData.depthType = GL_FLOAT;
+
+            _screenSpaceProcessor.draw(ssProcData);
+        }
```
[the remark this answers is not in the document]
tribal-tec: We agreed to remove the screen space processor.

brayns/Brayns.cpp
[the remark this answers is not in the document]
tribal-tec: No, it stores the buffers that are created in this file.

tribal-tec:
All should be good now

PR #140, opened 2017-03-29
Fixed morphology layout feature

favreau:

PR #617, opened 2018-11-27
Fix docker image

tribal-tec:


hernando:
As shown in HighFive #140 you may need to bump the HighFive version if you switch to boost >= 1.64

By the way, what's the reason for using 1.67? I hope this is not the minimum version now.

tribal-tec:
That's the default version for Debian Buster now. Tried to use stretch, but their assimp version is very old. I could force installing the 1.62 dev files if we have problems with 1.67.

PR #618, opened 2018-11-27
Removal of metaball and layout options from morphology loading.

hernando:
Arranging morphologies after loading can be easily done with a Python
scripts, as morphologies are imported as separate models if loaded from
a directory.

"circuit" and "morphology" prefixes stripped from loading parameters of
the CircuitViewer plugin.

"Mesh transformation" renamed to "transform meshes"

Refactorings.


plugins/CircuitViewer/io/CircuitLoader.cpp
```diff
@@ -129,74 +124,115 @@ struct CircuitProperties
...
-    int32_t circuitMeshTransformation = 0;
+    double density = 0.0;
+    int32_t randomSeed = 0;
+    std::string report;
+    std::vector<std::string> targetList;
+    std::string targets;
+    std::string meshFilenamePattern;
+    std::string meshFolder;
+    bool useSimulationModel = false;
+    bool transformMeshes = 0;
```
karjonas:
```suggestion
    bool transformMeshes = false;
```

plugins/CircuitViewer/io/CircuitLoader.cpp
```diff
@@ -688,7 +657,16 @@ ModelDescriptorPtr CircuitLoader::importFromFile(
...
     properties.merge(propertiesTmp);
     auto impl = Impl(_scene, properties);
-    return impl.importCircuit(filename, callback);
+    try
+    {
+        return impl.importCircuit(filename, callback);
+    }
+    catch (const std::exception& error)
+    {
+        BRAYNS_ERROR << "Failed to open " << filename << ": " << error.what()
```
karjonas: Why not let the exception throw? I think if you load from the web ui and this fails then you get no notification.

hernando: OK, the try catch was already there, but from what you say it's better to remove it.

brayns/Brayns.cpp
```diff
@@ -324,81 +324,55 @@ struct Brayns::Impl : public PluginAPI
         const GeometryParameters& geometryParameters =
```
tribal-tec: so next commit when this code will be moved to the CircuitViewerPlugin? :smile:

brayns/Brayns.cpp
```diff
@@ -324,81 +324,55 @@ struct Brayns::Impl : public PluginAPI
...
-
-            pm.setProperty({"morphologyLayoutHorizontalSpacing",
-                            "morphologyLayoutHorizontalSpacing",
-                            int32_t(layout.horizontalSpacing),
-                            {0, std::numeric_limits<int32_t>::max()}});
+            pm.addProperty({"boundingBoxP0", "boundingBoxP0", p0});
+            pm.addProperty({"boundingBoxP1", "boundingBoxP1", p1});
         }
-
+        pm.addProperty({"dampenBranchThicknessChangerate",
```
tribal-tec: `moreBeautifulBranches`? any name is probably terrible, do we need this option really?

brayns/Brayns.cpp
```diff
@@ -324,81 +324,55 @@ struct Brayns::Impl : public PluginAPI
...
+                         "dampenBranchThicknessChangerate",
+                         geometryParameters
+                             .getMorphologyDampenBranchThicknessChangerate()});
+        pm.addProperty({"useSDFGeometries", "useSDFGeometries",
+                         geometryParameters.getMorphologyUseSDFGeometries()});
+        pm.addProperty({"loadCacheFile", "loadCacheFile",
```
tribal-tec: cache still here?

brayns/Brayns.cpp
```diff
@@ -431,40 +405,26 @@ struct Brayns::Impl : public PluginAPI
...
-            pm.setProperty({"morphologySectionTypesSoma",
-                            "morphologySectionTypesSoma", soma});
-            pm.setProperty({"morphologySectionTypesAxon",
-                            "morphologySectionTypesAxon", axon});
-            pm.setProperty({"morphologySectionTypesDendrite",
-                            "morphologySectionTypesDendrite", dendrite});
-            pm.setProperty({"morphologySectionTypesApicalDendrite",
-                            "morphologySectionTypesApicalDendrite",
-                            apicalDendrite});
+            pm.addProperty({"sectionTypesSoma", "sectionTypesSoma", soma});
```
tribal-tec: what happend to 'fullDetailed', 'noAxon' and 'onlySomas'?

brayns/Brayns.cpp
```diff
@@ -431,40 +405,26 @@ struct Brayns::Impl : public PluginAPI
...
-                        int32_t(geometryParameters.getMetaballsGridSize())});
-        pm.setProperty({"metaballsThreshold", "metaballsThreshold",
-                        double(geometryParameters.getMetaballsThreshold())});
-        pm.setProperty(
-            {"metaballsSamplesFromSoma", "metaballsSamplesFromSoma",
-             int32_t(geometryParameters.getMetaballsSamplesFromSoma())});
-        pm.setProperty({"useRealisticSomas", "useRealisticSomas",
-                        geometryParameters.useRealisticSomas()});
-
+        pm.addProperty({"endSimulationTime", "endSimulationTime",
```
tribal-tec: remove those?

plugins/CircuitViewer/io/CircuitLoader.cpp
```diff
@@ -129,74 +124,115 @@ struct CircuitProperties
...
+CompartmentReportPtr _openCompartmentReport(const brion::BlueConfig& blueConfig,
+                                            const std::string& name,
+                                            const brion::GIDSet& gids)
+{
+    try
+    {
+        auto source = blueConfig.getReportSource(name);
+        if (source.getPath().empty())
+        {
+            BRAYNS_ERROR << "Compartment report not found: " << name
```
tribal-tec: as mentioned by @karjonas, errors shall be thrown for the client to get them. Local prints can be done if you wish, but the throw is actually more important.

brayns/Brayns.cpp
hernando: For later, this commit is older than this discussion and I need it merged for the upcoming ones.

brayns/Brayns.cpp
hernando: I already did it, but it was conflicting with Jonas' PR so I abandoned it. If you are interested in what I did, look at commits hernando/Brayns@eef533db and hernando/Brayns@e911400

tribal-tec: The plugin options we should discuss. They way I see it is that the PropertyMap can do all of that, also initializing them from commandline. The plugin can then define its own commandline options, and the core shouldn't know about them. Let's discuss once you're back.

brayns/Brayns.cpp
hernando: I would do it also later

plugins/CircuitViewer/io/CircuitLoader.cpp
hernando: Let me leave this code as is for this commit because it has changed in a follow up and I think that now it directly throws.

brayns/Brayns.cpp
hernando: Caching appears at multiple places. I prefer to have a dedicated PR for dealing with that.

brayns/Brayns.cpp
hernando: That's not what it means (which shows it's a bad name). I suggested lowPassSectionRadius, reduceRadiusVariation or similar, because that's what the option actually does. It's not reducing the change rate (samples stay where they are), but the difference between samples.

hernando:
Retest this please

hernando:
Retest this please

hernando:
ping

PR #622, opened 2018-11-30
Use Brain's object for simulation configs and reports.

hernando:
After this commit, no low level Brion objects are used anymore.
The color scheme neuron_by_layer now depends on the existence of
LayerX targets and not in MVD2.


plugins/CircuitViewer/io/CircuitLoader.cpp
```diff
@@ -107,6 +106,52 @@ std::string _getMeshFilenameFromGID(const uint64_t gid,
...
+ */
+size_ts _getLayerIds(const brain::Circuit& circuit, const brain::GIDSet& gids)
+{
+    std::vector<brion::GIDSet> layers;
+    try
+    {
+        for (auto layer : {"1", "2", "3", "4", "5", "6"})
+            layers.push_back(circuit.getGIDs(std::string("Layer") + layer));
+    }
+    catch (...)
```
karjonas: Can you specify what you expect to catch here?

plugins/CircuitViewer/io/CircuitLoader.cpp
```diff
@@ -107,6 +106,52 @@ std::string _getMeshFilenameFromGID(const uint64_t gid,
...
+    size_ts materialIds;
+    materialIds.reserve(gids.size());
+    // For each GID find in which layer GID list it appears, assign
+    // this index to its material id list.
+    for (auto gid : gids)
+    {
+        size_t i = 0;
+        for (i = 0; i != layers.size(); ++i)
+        {
+            while (iters[i] != layers[i].end() && *iters[i] < gid)
```
karjonas: Is there a reason why you store iterators and use while loops instead of using for-loop or std::find for each layer?

plugins/CircuitViewer/io/CircuitLoader.cpp
```diff
@@ -182,13 +227,12 @@ struct CircuitProperties
...
-                               const CircuitProperties& properties)
+brain::GIDSet _getFilteredGIDs(
+    const std::function<brion::GIDSet(const std::string&)>& resolver,
+    const brain::Circuit& circuit, const std::string& target,
+    const CircuitProperties& properties)
 {
-    const auto allGIDs = circuit.getRandomGIDs(properties.density / 100.0,
-                                               target, properties.randomSeed);
-
+    const auto allGIDs = resolver(target);
```
karjonas: I find giving `resolver` and `target` as input a bit strange. What about giving a `brion::GidSet&& allGIDs` as input instead?

plugins/CircuitViewer/io/CircuitLoader.cpp
```diff
@@ -670,7 +700,7 @@ std::string CircuitLoader::getName() const
 
 std::vector<std::string> CircuitLoader::getSupportedExtensions() const
 {
-    return {"BlueConfig", "BlueConfig3", "CircuitConfig", "circuit"};
+    return {"BlueConfig", "BlueConfig3", "CircuitConfig", ".json", "circuit"};
```
karjonas: Can you make this a common constant variable with the `names` variable above?

karjonas:
Can you clarify this part of the commit message "The color scheme neuron_by_layer now depends on the existence of LayerX targets and not in MVD2."?

plugins/CircuitViewer/io/CircuitLoader.cpp
hernando: Yes, based on the fact that GID sets are sorted, this code finds in which layer is present each gid by visiting exactly only once each element of every set. What you are proposing would be noticeably slower for milliions of cells (let's say O(n log n) vs O(n))

plugins/CircuitViewer/io/CircuitLoader.cpp
hernando: Notice that the resolver may be a brain::Simulation or a brain::Circuit depending on which type of config file was used for loading. This is not the case with BlueConfigs, but it's for SONATA.
Anyway, I have removed this filtering in a follow up commit, so I wouldn't pay much attention to this code at this point.

plugins/CircuitViewer/io/CircuitLoader.cpp
karjonas: Can you add the above comment in the code as well?

plugins/CircuitViewer/io/CircuitLoader.cpp
```diff
@@ -634,7 +667,7 @@ bool CircuitLoader::isSupported(const std::string& filename,
     };
 
     const std::set<std::string> names = {"BlueConfig", "BlueConfig3",
-                                         "CircuitConfig", "circuit"};
+                                         "CircuitConfig", ".json", "circuit"};
 
     for (const auto& name : names)
```
karjonas:
```suggestion
    for (const auto& name : LOADER_EXTENSIONS)
```

plugins/CircuitViewer/io/CircuitLoader.cpp
hernando: Done

plugins/CircuitViewer/io/CircuitLoader.cpp
```diff
@@ -261,23 +299,42 @@ class Impl
...
+        // Open circuit and simulation if possible
+        brion::URI uri(source);
+        std::unique_ptr<brain::Simulation> simulation;
+        try
+        {
+            simulation = std::make_unique<brain::Simulation>(uri);
+        }
+        catch (std::runtime_error& error)
+        {
+            BRAYNS_INFO << "Cannot open " + source + " as simulation config: "
```
tribal-tec: This message is printed always or only when report is requested to load?

plugins/CircuitViewer/io/CircuitLoader.cpp
```diff
@@ -261,23 +299,42 @@ class Impl
...
+        const strings localTargets = _properties.targetList.empty()
+                                         ? strings{{""}}
+                                         : _properties.targetList;
+
+        auto resolver = [&](const std::string& name) {
+            const float fraction = _properties.density / 100.0f;
+            if (simulation)
+                return simulation->getGIDs(name, fraction,
+                                           _properties.randomSeed);
+            return circuit.getRandomGIDs(fraction, name,
```
tribal-tec: nice that the signature is not the same for those two getID functions :stuck_out_tongue:

plugins/CircuitViewer/io/MorphologyLoader.h
```diff
@@ -25,6 +25,7 @@
 #include <brayns/common/types.h>
 #include <brayns/parameters/GeometryParameters.h>
 
+#include <brain/compartmentReportMapping.h>
```
tribal-tec: forward declare should be enough

plugins/CircuitViewer/io/CircuitLoader.cpp
tribal-tec: I was referring to `name, fraction` vs `fraction, name` argument swap. Asking the simulation vs the circuit is also a valid point.

plugins/CircuitViewer/io/CircuitLoader.cpp
hernando: It's unrelated to report and it will never be printed with BlueConfigs. This code deals with the fact that in SONATA circuit and simulation configs are different things (hooray!) and we want to support both.

plugins/CircuitViewer/io/MorphologyLoader.h
```diff
@@ -35,6 +35,11 @@ namespace servus
 class URI;
 }
 
+namespace brain
```
tribal-tec: there is no `<brain/types.h>`?

plugins/CircuitViewer/io/CircuitLoader.cpp
```diff
@@ -261,23 +299,42 @@ class Impl
...
-        const brion::BlueConfig bc(source);
-        const brain::Circuit circuit(bc);
+        // Open circuit and simulation if possible
+        brion::URI uri(source);
+        std::unique_ptr<brain::Simulation> simulation;
+        try
+        {
+            simulation = std::make_unique<brain::Simulation>(uri);
+        }
+        catch (std::runtime_error& error)
```
tribal-tec: `const&`

plugins/CircuitViewer/io/MorphologyLoader.h
hernando: And Servus/types.h for that matter.

PR #624, opened 2018-12-03
Fix function signature and missing autocomplete in generated python functions

tribal-tec:


python/brayns/utils.py
```diff
@@ -172,6 +172,7 @@ def add_progress_cancel_widget(func):  # pragma: no cover
     :return: the decorator
     :rtype: decorator
     """
+    @wraps(func)
```
hernando: Is this decorator on par with the one provided by the decorator module?

PR #626, opened 2018-12-05
DeflectPlugin improvements

tribal-tec:


plugins/Deflect/utils.h
```diff
@@ -0,0 +1,60 @@
...
+ * along with this library; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#pragma once
+
+#include <brayns/common/utils/enumUtils.h>
+#include <deflect/deflect.h>
+
+#define deflectPixelOp deflectPixelOp
```
karjonas: I don't understand what this does

plugins/Deflect/utils.h
```diff
@@ -0,0 +1,60 @@
...
+            return deflect::View::right_eye;
+        return deflect::View::mono;
+    }
+    return deflect::View::mono;
+}
+
+inline uint8_t getChannel(const std::string& name)
+{
+    if (name.length() == 2)
+        return std::atoi(&name.at(0));
```
karjonas:
```suggestion
        return std::stoi(name);
```

brayns/Brayns.cpp
```diff
@@ -257,14 +256,13 @@ struct Brayns::Impl : public PluginAPI
     {
         _engine.reset(); // Free resources before creating a new engine
 
-        const auto& engineName =
-            _parametersManager.getApplicationParameters().getEngine();
+        auto& appParams = _parametersManager.getApplicationParameters();
+
+        const auto& engineName = appParams.getEngine();
```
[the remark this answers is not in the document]
hernando: This is in fact misleading, I thought the variable type was a string until the code below made me double check and I discovered it's an enum. In fact, I think that EngineType should go away as plugins should be able to provide additional engines.

brayns/PluginManager.cpp
```diff
@@ -43,6 +40,11 @@ typedef ExtensionPlugin* (*CreateFuncType)(int, const char**);
 
 PluginManager::PluginManager(int argc, const char** argv)
 {
+    bool help = false;
+    for (int i = 0; i < argc && !help; ++i)
+        if (std::string(argv[i]) == "--help")
+            help = true;
```
hernando: const bool help = std::count_if(argv, argv + argc, [](const char* arg) {
        return std::strcmp(*arg, "--help") == 0;
    }) > 0;

brayns/PluginManager.cpp
```diff
@@ -137,10 +129,12 @@ void PluginManager::_loadPlugin(const char* name, int argc, const char* argv[])
         }
 
         CreateFuncType createFunc = (CreateFuncType)createSym;
-
-        _extensions.emplace_back(createFunc(argc, argv));
-        _libs.push_back(std::move(library));
-        BRAYNS_INFO << "Loaded plugin '" << name << "'" << std::endl;
+        if (auto plugin = createFunc(argc, argv))
```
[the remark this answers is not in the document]
hernando: I would delegate that to the createFunc itself.

brayns/common/PropertyMap.cpp
```diff
@@ -212,7 +212,7 @@ void PropertyMap::merge(const PropertyMap& input)
         };
     };
 
-    const auto isCompatbileTypes = [](Property::Type t0, Property::Type t1) {
+    const auto areCompatibleTypes = [](Property::Type t0, Property::Type t1) {
```
hernando: const auto canBePromoted = [](Property::Type from, Property::Type to)

From the name is not clear in which order the conversion is allowed.

brayns/Brayns.cpp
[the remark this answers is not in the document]
tribal-tec: I'll revert that change as this is not relevant at all for this PR.

brayns/PluginManager.cpp
tribal-tec: lol, shame on the original author of this in ParametersManager.cpp :)

plugins/Deflect/utils.h
tribal-tec: It defines a value to TEXTIFY it. The name for `OSP_REGISTER_PIXEL_OP` is just this 'name', but for creating it at runtime I need a string for that. Hence this magic.

plugins/Rockets/RocketsPlugin.cpp
```diff
@@ -294,16 +286,20 @@ class RocketsPlugin::Impl : public ActionInterface
...
         ](const auto& request) {
-            PropertyMap params = input;
-            if (::from_json(params, request.message))
+            engine.triggerRender();
+            try
             {
-                engine.triggerRender();
-                return Response{to_json(action(params))};
+                return Response{
+                    to_json(action(jsonToPropertyMap(request.message)))};
```
hernando: Is the code truly equivalent? I mean, before action was invoked with default values coming from input, but now those values are gone. Doesn't this have any unexpected side effect?

plugins/Rockets/RocketsPlugin.cpp
```diff
@@ -294,16 +286,20 @@ class RocketsPlugin::Impl : public ActionInterface
                          const std::function<PropertyMap(PropertyMap)>& action)
     {
         _bindEndpoint(desc.methodName, [
-            name = desc.methodName, input, action, &engine = _engine
+            name = desc.methodName, action, &engine = _engine
         ](const auto& request) {
-            PropertyMap params = input;
-            if (::from_json(params, request.message))
+            engine.triggerRender();
```
hernando: Why do you need this? It doesn't look right to trigger rendering from any possible request.

plugins/Deflect/utils.h
```diff
@@ -0,0 +1,60 @@
...
+template <>
+inline std::vector<std::pair<std::string, deflect::ChromaSubsampling>> enumMap()
+{
+    return {{"yuv444", deflect::ChromaSubsampling::YUV444},
+            {"yuv422", deflect::ChromaSubsampling::YUV422},
+            {"yuv420", deflect::ChromaSubsampling::YUV420}};
+}
+namespace utils
+{
+inline deflect::View getView(const std::string& name)
```
hernando: Document somewhere the assumptions of what name encodes

plugins/Deflect/DeflectParameters.h
```diff
@@ -0,0 +1,148 @@
...
+    unsigned getQuality() const
+    {
+        return (unsigned)_props.getProperty<int32_t>(PARAM_QUALITY);
+    }
+    void setQuality(const unsigned quality)
+    {
+        _updateProperty(PARAM_QUALITY, (int32_t)quality);
+    }
+
+    /** Stream ID; defaults to DEFLECT_ID if empty */
```
hernando: I don't understand this comment. Is it an instruction to the consumer?

plugins/Deflect/DeflectParameters.cpp
```diff
@@ -0,0 +1,76 @@
...
+{
+    PropertyMap properties("Deflect plugin parameters");
+    properties.setProperty(
+        {PARAM_ENABLED, true,
+         Property::MetaData{"Enable streaming", "Enable/disable streaming"}});
+    properties.setProperty(
+        {PARAM_ID, std::string(),
+         Property::MetaData{
+             "Stream ID",
+             "The ID/name of the stream, equivalent to DEFLECT_ID"}});
```
hernando: What is DEFLECT_ID, an env var? I think you're assume too much knowledge from the reader's point of view.

plugins/Deflect/DeflectParameters.cpp
```diff
@@ -0,0 +1,76 @@
...
+    properties.setProperty(
+        {PARAM_HOSTNAME, std::string(),
+         Property::MetaData{"Stream hostname", "Hostname of Deflect server"}});
+    properties.setProperty(
+        {PARAM_PORT,
+         (int32_t)deflect::Stream::defaultPortNumber,
+         {1023, 65535},
```
hernando: The range should start at 1, I don't see any reason to prevent running Deflect server on a privileged port (which doesn't necessarily imply it runs as a privileged process)

plugins/Deflect/DeflectParameters.cpp
```diff
@@ -0,0 +1,76 @@
...
+         Property::MetaData{"Stream port", "Port of Deflect server"}});
+    properties.setProperty(
+        {PARAM_COMPRESSION, true,
+         Property::MetaData{"Use JPEG compression", "Use JPEG compression"}});
+    properties.setProperty(
+        {PARAM_TOP_DOWN, false,
+         Property::MetaData{
+             "Stream image top-down",
+             "Top-down image orientation instead of bottom-up"}});
```
hernando: FLIP_IMAGE?

plugins/Deflect/DeflectPixelOp.cpp
```diff
@@ -19,16 +19,18 @@
...
-template <typename T>
-std::future<T> make_ready_future(const T value)
+std::future<bool> make_ready_future()
 {
-    std::promise<T> promise;
-    promise.set_value(value);
+    std::promise<bool> promise;
+    promise.set_value(true);
     return promise.get_future();
 }
```
hernando: I remember mentioning in another PR that this function was not needed, did you do any changes back then?

plugins/Deflect/DeflectPlugin.cpp
```diff
@@ -81,31 +81,32 @@ class DeflectPlugin::Impl
                 _closeStream();
         }
 
-        if (_previousHost != _params.getHostname())
-        {
-            _params.setEnabled(true);
-            _previousHost = _params.getHostname();
-        }
-
         const bool deflectEnabled = _params.getEnabled();
```
hernando: add !_params.getEnabled() to the if (_stream) statement.
Check for _params.getEnabled() here and if false _params.resetModified() and early return. You don't need the deflectEnabled variable afterwards.

plugins/Deflect/DeflectPlugin.cpp
```diff
@@ -81,31 +81,32 @@ class DeflectPlugin::Impl
...
         if (_stream && _stream->isConnected() && !deflectEnabled)
```
hernando: This statement is not needed with the change suggested above as the stream would have been closed in the first if statement.

plugins/Deflect/DeflectPlugin.cpp
```diff
@@ -81,31 +81,32 @@ class DeflectPlugin::Impl
...
             _closeStream();
 
-        const bool observerOnly = _engine.haveDeflectPixelOp();
+        const bool observerOnly = _params.usePixelOp();
         if (deflectEnabled && !_stream && _startStream(observerOnly))
```
hernando: observerOnly parameter not needed as _startStream has access to _params (and in facts uses them)

plugins/Deflect/DeflectParameters.cpp
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+#include "DeflectParameters.h"
+#include "utils.h"
+
+#include <deflect/Stream.h>
+
+namespace brayns
+{
+PropertyMap DeflectParameters::createPropertyMap()
```
hernando: createDefaultProperties. getDefaultProperties?

plugins/Deflect/DeflectPlugin.cpp
```diff
@@ -449,3 +458,12 @@ void DeflectPlugin::postRender()
...
 }
 }
+
+extern "C" brayns::ExtensionPlugin* brayns_plugin_create(const int argc,
+                                                         const char** argv)
+{
+    auto pm = brayns::DeflectParameters::createPropertyMap();
+    if (!pm.parse(argc, argv))
+        return nullptr;
+    return new brayns::DeflectPlugin(pm);
```
hernando: Better:

    brayns::DeflectParameters params;
    if (!params.parse(argc, argv))
        return nullptr;
    return new brayns::DeflectPlugin(std::move(params));

That way createPropertyMap can become an implementation function and DeflectPlugin takes directly what it needs, which is the parameters.

engines/ospray/OSPRayRenderer.cpp
```diff
@@ -170,8 +170,8 @@ void OSPRayRenderer::_createOSPRenderer()
 
 void OSPRayRenderer::_commitRendererMaterials()
 {
-    _scene->visitModels([&renderer = _currentOSPRenderer](Model& model) {
-            static_cast<OSPRayModel&>(model).commitMaterials(renderer);
+    _scene->visitModels([& renderer = _currentOSPRenderer](Model & model) {
+        static_cast<OSPRayModel&>(model).commitMaterials(renderer);
```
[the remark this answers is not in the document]
tribal-tec: clang-format. dunno

plugins/Rockets/RocketsPlugin.cpp
tribal-tec: It was actually wrong before :)

plugins/Rockets/RocketsPlugin.cpp
tribal-tec: maybe not. I have to check

plugins/Deflect/utils.h
tribal-tec: the code is enough documentation, no?

plugins/Deflect/DeflectParameters.h
tribal-tec: it's an explanation what the value is if not specified.

plugins/Deflect/DeflectParameters.cpp
tribal-tec: If you don't know deflect or tide, using this probably isn't advised anyways.

plugins/Deflect/DeflectParameters.cpp
tribal-tec: fair enough

plugins/Deflect/DeflectParameters.cpp
tribal-tec: then it's less obvious what orientation it has?!

plugins/Deflect/DeflectPixelOp.cpp
tribal-tec: the function is needed, no? I don't remember what the solution for the comment the other day was.

plugins/Deflect/utils.h
hernando: But where are these names coming from? My fear is that the knowledge is distributed in multiple places, making maintenance more complicated.

plugins/Deflect/DeflectPlugin.cpp
tribal-tec: I'm lost. Apparently you are too from the current code. We should look at this in person.

brayns/common/PropertyMap.cpp
```diff
@@ -212,7 +212,7 @@ void PropertyMap::merge(const PropertyMap& input)
...
         return (t0 == Property::Type::Int && t1 == Property::Type::Double) ||
                (t0 == Property::Type::Double && t1 == Property::Type::Int) ||
```
hernando: This conversion should not be allowed @karjonas, any reason for allowing it?

karjonas: When we get doubles from the UI they are sent as `0` instead of `0.0` so therefore we need to be able to convert between double and float.

hernando: But that would be a type promotion, which is fine. The condition is also accepting type demotion (accepting double for an integer), which is what I don't see why should be necessary.

karjonas: Yes, should be fine, you just need to change the order in the jsonToPropertyMap method in jsonPropertyMap.h.

PR #627, opened 2018-12-05
Morphology loading refactorings

hernando:
ParallelModelContainer renamed to ModelData.

Bugfixes in SDF morphology generation:
- Spurious sphere at terminal sections
- Missing smooth bifurcations when radius difference dampening was
  not enabled.

Removed unnecessary code and SDF model creation optimized.


plugins/CircuitViewer/io/MorphologyLoader.cpp
```diff
@@ -87,10 +89,202 @@ const Property PROP_GEOMETRY_QUALITY = {
...
+    case brain::neuron::SectionType::axon:
+        return mask & static_cast<size_t>(MorphologySectionType::axon);
+    case brain::neuron::SectionType::dendrite:
+        return mask & static_cast<size_t>(MorphologySectionType::dendrite);
+    case brain::neuron::SectionType::apicalDendrite:
+        return mask &
+               static_cast<size_t>(MorphologySectionType::apical_dendrite);
+    case brain::neuron::SectionType::undefined:
+    default:
+        return true; // What do we do here?
```
tribal-tec: return false? throw? true seems wrong, but I would have to follow where this is called from.

hernando: I was thinking that removing things that we don't know about it's probably not the most desirable behaviour. I was having astrocytes in mind when I wrote this code.

plugins/CircuitViewer/io/MorphologyLoader.cpp
```diff
@@ -87,10 +90,214 @@ const Property PROP_GEOMETRY_QUALITY = {
...
+    case brain::neuron::SectionType::undefined:
+    default:
+        return true; // Unclassified sections will always be visible.
+    }
 }
 
-namespace brayns
+inline float _getLastSampleRadius(const brain::neuron::Section& section)
+{
+    return section[-1][3] * 0.5;
```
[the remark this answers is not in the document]
hernando: Yes it's safe. Section::operator[] borrows its semantics from Python.

PR #628, opened 2018-12-06
Fix and move the Perspective Parallax Camera from modules to core

chevtche:


chevtche:
We already discussed with Daniel. The original camera use a weird algorithm that I don't understand. I don't know how to add the parrallax plane to this one without breaking the other fancy stuff (ex: aperture). It's why kept it as it is and added a new one.

tribal-tec:
I agree with @hernando that ideally the one perspective camera would also support the zero parallax plane. Maybe this can be still figured out how to make this work with the aperture?

chevtche:
Ok. Let's look at the code together. Maybe 3 of us can understand it.

PR #629, opened 2018-12-06
Set default transfer function and renderer for circuit simulations

tribal-tec:


plugins/CircuitViewer/io/CircuitLoader.cpp
```diff
@@ -419,6 +559,9 @@ class Impl
             // Only keep GIDs from the report
             allGids = compartmentReport->getGIDs();
             reportMapping = &compartmentReport->getMapping();
+
+            _setSimulationTransferFunction(model->getTransferFunction());
+            _renderer.setCurrentRenderer("basic_simulation");
```
hernando: I would not do this here, because it will change your renderer every time you load a circuit, which I probably is surprising. I would move it instead to the CircuitViewer::init method instead

tribal-tec: The same applies when loading the plugin then but not loading a circuit, yet the renderer gets changed. Same difference for me. My solution at least changes the renderer only if there is an actual simulation.

tribal-tec:
updated

PR #632, opened 2018-12-07
Refactorings

hernando:


plugins/CircuitViewer/io/CircuitLoader.cpp
```diff
@@ -117,6 +114,107 @@ std::string _getMeshFilenameFromGID(const uint64_t gid,
...
+    radii["L6_NBC"] = 9.33885664259;
+    radii["L6_SBC"] = 8.06322860718;
+    radii["L6_ChC"] = 8.15173133214;
+
+    const auto& mtypes = circuit.getMorphologyTypeNames();
+    std::vector<float> output;
+    for (auto mtype : mtypes)
+    {
+        auto iter = radii.find(mtype);
+        if (iter == radii.end())
```
tribal-tec: ideally that doesn't happen, log when it does?

plugins/CircuitViewer/io/CircuitLoader.cpp
```diff
@@ -584,72 +672,75 @@ class Impl
...
+
+                auto data =
+                    morphLoader.processMorphology(*morphology, morphologyIndex,
+                                                  materialFunc, reportMapping,
+                                                  _morphologyParams);
+#pragma omp critical
+                data.dump(model);
 
+                ++current;
+                // Since throwing from inside an parallel for is not a good
```
tribal-tec:
```suggestion
                // Throwing (happens if loading is cancelled) from inside a parallel-for is not allowed.
```

plugins/CircuitViewer/io/CircuitLoader.cpp
```diff
@@ -584,72 +672,75 @@ class Impl
...
-#pragma omp atomic
-                        ++loadingFailures;
-#pragma omp critical
-                    modelData.addTo(model);
+                                            current / float(gids.size()));
                 }
                 catch (...)
                 {
                     cancelException = std::current_exception();
-                    morphologyIndex = uris.size();
```
tribal-tec: The loading then continues until `uris.size()`, even though it was cancelled?!

plugins/CircuitViewer/io/CircuitLoader.cpp
hernando: Yes, for mtypes not in the list above (e.g. hippocampus). Do you want to see a warning printed for every cell for which its mytpe is not found or just once per mtype?

tribal-tec: Only once at the end that there were mtypes (maybe list them there) that were not in this list.

plugins/CircuitViewer/io/CircuitLoader.cpp
hernando: There was something missing in the outer loop to stop the iteration if the cancel exception is raised.

hernando:
Depends on Brion#252

PR #636, opened 2018-12-11
Fix docker build, other fixes

tribal-tec:


CMakeLists.txt
```diff
@@ -81,7 +81,7 @@ if(ROCKETS_FOUND AND ROCKETS_USE_LIBWEBSOCKETS)
     list(APPEND COMMON_FIND_PACKAGE_DEFINES BRAYNS_USE_NETWORKING)
   endif()
 else()
-  unset(BRAYNS_NETWORKING_ENABLED)
+  set(BRAYNS_NETWORKING_ENABLED OFF)
```
hernando: `set(BRAYNS_NETWORKING_ENABLED)` should also work, doesn't it?

tribal-tec: Don't know, maybe? The explicit one is more obvious

tribal-tec: According to documentation it has the semantics as `unset`, hence the variable is uninitialized. But as it was set to `ON` before, this has no effect in practice, only an explicit set to `OFF`.

tribal-tec:
retest this please

ppodhajski:
retest this please

ppodhajski:
retest this please

tribal-tec:
retest this please

tribal-tec:
ping

PR #638, opened 2018-12-12
Simplify plugin definition where defining `init()` becomes optional

tribal-tec:
The `_api` becomes a member of each plugins' base class, and it's safe to use
directly within `init()`.


tribal-tec:
There is no more necessity to implement `init()`. The pluginAPI is needed in all plugins, so it's a simplification.

PR #640, opened 2018-12-13
Reduced likelihood of ambiguous overload in Property constructors

hernando:


hernando:
Needs approval again, I hadn't compiled the VRPN plugin.

brayns/common/PropertyMap.h
```diff
@@ -247,6 +247,22 @@ class PropertyMap
     {
     }
 
+    PropertyMap(const PropertyMap& other) = default;
+
+    // std::vector move constructor is not noexcept until C++17, if we want
+    // this class to be movable we have to do it by hand.
+    PropertyMap(PropertyMap&& other) noexcept
```
tribal-tec: there is no test for move & copy assignment & construction. This is not needed right now I guess, but you need that for later?

hernando: I wanted to have a move constructor and assignment, but copy assignment is already being used. Move construction could also have been used without this, but due to the issue in the comment above, the default move constructor was being deleted (apart from the fact there's a explicit declaration of the copy constructor). As you see, we are not 
Do you really what a test for this? The code is pretty trivial, will that be testing Brayns, the compiler or be a C++11 quiz? (as you can see I'm not an advocate of 100% code coverage)

hernando:
I have another PR on top of this one prepared, do you any other comments?

PR #641, opened 2018-12-13
Add the ability to scale the Opendeck camera

chevtche:


plugins/OpenDeck/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
-# Copyright (c) 2015-2018, EPFL/Blue Brain Project
-# All rights reserved. Do not distribute without permission.
+#Copyright(c) 2015 - 2018, EPFL / Blue Brain Project
```
tribal-tec: do not clang-format CMake files

plugins/OpenDeck/OpenDeckPlugin.cpp
```diff
@@ -38,58 +38,62 @@ constexpr char leftWallBufferName[] = "0L";
...
-                               const Vector2ui& floorRes)
-    : _wallRes(wallRes)
-    , _floorRes(floorRes)
+OpenDeckPlugin::OpenDeckPlugin(OpenDeckParameters&& params)
+    : _params(std::move(params))
 {
+    if (_params.getResolutionScaling() > 1.0f ||
+        _params.getResolutionScaling() <= 0.0f)
+    {
+        throw std::runtime_error(
```
tribal-tec: that has to be catched in the C create() function, no exceptions are allowed to pass through there

PR #643, opened 2018-12-13
Added a method to merge only known properties from one map to another

hernando:


brayns/common/PropertyMap.h
```diff
@@ -355,6 +363,9 @@ class PropertyMap
     const auto& getProperties() const { return _properties; }
     /** Merge this property map with properties from another. */
     void merge(const PropertyMap& input);
+    /** Take the values from another property map only for properties that
+        are known and compatible to this one. */
```
tribal-tec: if a property is incompatible it throws. Shall be mentioned for this and for `merge`

hernando: That happens in both cases.

brayns/common/PropertyMap.h
```diff
@@ -353,9 +358,19 @@ class PropertyMap
...
+     *  incompatible types is found in both maps.
+     */
     void merge(const PropertyMap& input);
 
+    /** Take the values from another property map only for properties that
+     *  are known and compatible to this one.
+     *  @throw std::runtime error if a property with the same name but
+     *  incompatible types is found in both maps.
+     */
+    void take(const PropertyMap& input);
```
karjonas: Maybe we should call these takeAll and takeExisting to make the difference between the two methods clearer. They only differ in an else clause anyway.

hernando: I prefer updateExisting rather than takeExisting and leave merge as merge.

PR #644, opened 2018-12-14
Add `--play-animation` switch to enable initial simulation playback for demos

tribal-tec:


brayns/parameters/AnimationParameters.cpp
```diff
@@ -22,23 +22,27 @@
...
 namespace brayns
 {
 AnimationParameters::AnimationParameters()
     : AbstractParameters("Animation")
 {
-    _parameters.add_options()(PARAM_ANIMATION_FRAME.c_str(),
-                              po::value<uint32_t>(),
-                              "Scene animation frame [float]");
+    _parameters.add_options()(PARAM_ANIMATION_FRAME, po::value<uint32_t>(),
+                              "Scene animation frame [float]")(
```
karjonas:
```suggestion
                              "Scene animation frame [int]")(
```

PR #647, opened 2018-12-17
Removals

hernando:


plugins/CircuitViewer/io/MorphologyLoader.h
```diff
@@ -39,7 +39,6 @@ enum class NeuronColorScheme
 {
     none = 0,
     by_id = 1,
-    by_type = 2,
     by_segment_type = 3,
```
tribal-tec: keep the '2' gap or re-enumerate here?

hernando: Re-enumerate

PR #648, opened 2018-12-18
Move engine-specific files to new library `braynsEngine`

tribal-tec:
* files moved from brayns/common to brayns/engine
* engines can be loaded dynamically, similar to plugins
* only one engine is loaded once at startup, using `--engine`
* `--engine optix` still works, otherwise OSPRay will be used by default
* add `--verbose` flag to toggle parameters printing
* new library `braynsManipulators` to solve cyclic dependency core <-> engine


brayns/engine/Model.h
```diff
@@ -200,13 +199,37 @@ class ModelDescriptor : public ModelParams
...
 
-    BRAYNS_API virtual ~Model() = default;
+    /** Factory method to create an engine-specific material. */
+    BRAYNS_API virtual MaterialPtr createMaterial(const size_t materialId,
+                                                  const std::string& name) = 0;
 
-    virtual void commitGeometry() = 0;
+    /**
+     * Create a volume with the given dimensions, voxel spacing and data type
+     * where the are voxels are set via setVoxels() from any memory location.
```
karjonas:
```suggestion
     * where the voxels are set via setVoxels() from any memory location.
```

PR #649, opened 2018-12-18
Implemented target string parsing

hernando:
A target key can be a number >= 0, a range given by to hyphen
separated numbers or a target name.


hernando:
Any recommendations about how to add tests? (not that it's not tested)
I'm leaning to make the added functions part of an header file to be included directly by a unit test.

plugins/CircuitViewer/io/utils.h
```diff
@@ -0,0 +1,100 @@
...
+#include <brain/types.h>
+
+#include <random>
+#include <regex>
+
+namespace brayns
+{
+namespace
+{
+inline void _trim(std::string& s)
```
tribal-tec: or `boost::trim`, like in PluginManager

PR #653, opened 2018-12-19
Fixed public headers

favreau:


favreau:
I had to add this to be able to use the engine API. This is because the FrameBuffer class includes it, so it has to be deployed. I'll add FreeImage then. Thx.

PR #654, opened 2018-12-21
Generalized camera registration like it is done for renderers

tribal-tec:


brayns/engine/Engine.h
```diff
@@ -145,6 +145,13 @@ class Engine
     void addRenderer(const std::string& name,
                      const PropertyMap& properties = {});
 
+    /**
+     * Add a new camera with optional properties. The camera registration for a
+     * concrete engine is specific to the actual engine, e.g.
+     * OSP_REGISTER_CAMERA for OSPRay.
+     */
+    void addCamera(const std::string& name, const PropertyMap& properties = {});
```
hernando: Both this name and addRenderer can be confusing. You are not adding a camera, you are adding a camera type. At least the documentation should make it clear. Otherwise somebody could think that can use this method to render a scene from multiple viewpoints.

engines/ospray/OSPRayEngine.cpp
```diff
@@ -262,44 +262,39 @@ void OSPRayEngine::_createCameras()
                            0.0635,
                            {"Eye separation"}};
 
-    RenderingParameters& rp = _parametersManager.getRenderingParameters();
```
hernando: This was awkward indeed.

brayns/engine/Engine.h
tribal-tec: You are the one who would think that :stuck_out_tongue: I can rename it to `addCameraType`.

PR #659, opened 2019-01-11
Fixed OSPRay material commit

favreau:


engines/ospray/OSPRayMaterial.cpp
```diff
@@ -98,7 +98,8 @@ void OSPRayMaterial::commit(const std::string& renderer)
 {
     if (!isModified())
         return;
-    ospRelease(_ospMaterial);
```
tribal-tec: ospray checks for that already, so this does not make any difference!?

favreau: It does for me, that's why I made that change. I tested my code using a plugin that defines another default material class. And this somehow made the ospRelease line raise an exception. I can investiguate more but the fix definitely solved the problem :-/

tribal-tec: The 'problem' is that no other ospRelease within Brayns has this check, so it becomes inconsistent. You have an older ospray version maybe?

favreau: I understand, and agree. But I am not sure what to do there. I am using the 1.7.3 OSPRay release, and I get the same behaviour on both my machine and the cluster, so I don't think it's a build problem. We can discuss on Monday if you want. There is something fishy there in any case.

tribal-tec: Fishy indeed. I don't see why repeating a `if()` check in one library or another one makes any difference. Let's see next week.

PR #661, opened 2019-01-16
Fix missing ospray modules in MPI mode

tribal-tec:


tribal-tec:
Commit message updated

PR #662, opened 2019-01-17
Bugfixes

tribal-tec:


plugins/Rockets/RocketsPlugin.cpp
```diff
@@ -1492,7 +1498,8 @@ class RocketsPlugin::Impl : public ActionInterface
         const auto notifyEndpoint = getNotificationEndpointName(endpoint);
 
         _jsonrpcServer->bind<PropertyMap>(requestEndpoint, [& object = object] {
-            return object.getPropertyMap();
+            return object.hasProperties() ? object.getPropertyMap()
+                                          : PropertyMap();
```
hernando: What about moving the logic to `PropertyObject`?
It's not unreasonable to think as well that when the current type is changed an empty `PropertyMap` should be assigned to `_properties[_currentType]` if there isn't any. That would allow to replace `at()` with [] at a few places and avoid unexpected exceptions which are not documented (e.g. for `hasProperty`). `hasProperties` may even not be necessary, but if it's, the implementation would change to `!_properties[_currentyType].empty()`

tribal-tec: Good point. I was hesitating at first to do something like this but was too conservative with the fix in the end.

hernando: Better

PR #663, opened 2019-01-18
Bugfixes round 2

tribal-tec:


brayns/tasks/AddModelFromBlobTask.cpp
```diff
@@ -40,10 +40,11 @@ AddModelFromBlobTask::AddModelFromBlobTask(const BinaryParam& param,
...
     LoadModelFunctor functor{engine, param};
     functor.setCancelToken(_cancelToken);
-    functor.setProgressFunc([& progress = progress](const auto& msg, auto,
-                                                    auto amount) {
-        progress.update(msg, amount);
-    });
+    functor.setProgressFunc(
+        [& progress = progress, w = CHUNK_PROGRESS_WEIGHT ](const auto& msg,
+                                                            auto, auto amount) {
+            progress.update(msg, w + (amount * (1.f - w)));
```
hernando: A minor comment. I'm starting to think that the progress callback should not be a single progress bar, but a bar per step and there should be a separate method to signal when the process is done instead of relying on reaching 1. With this code, if you upload a huge mesh to the server, transmission accounts for half the progress, creating the mesh is 1/4 and the remaining 1/4 is postprocessing. I don't have numbers but it post processing happens to be slower than anything else, the progress bar becomes a bit misleading.

python/brayns/async_client.py
```diff
@@ -73,8 +73,7 @@ def image(self, size, format='jpg', animation_parameters=None, camera=None, qual
         :param int quality: compression quality between 1 (worst) and 100 (best)
         :param object renderer: renderer to use instead of current renderer
         :param int samples_per_pixel: samples per pixel to increase render quality
-        :return: the PIL image of the current rendering, None on error obtaining the image
-        :rtype: :py:class:`~PIL.Image.Image`
+        :return: the asyncio.Future of the snapshot generation, or None if inside a notebook
```
hernando: rtype gone?

engines/ospray/ispc/render/utils/SkyBox.ispc
```diff
@@ -66,9 +66,9 @@ vec4f skyboxMapping(const uniform Renderer* uniform renderer,
 
     // Only st needs to be set to sample from Texture2D
     varying DifferentialGeometry dg;
-    dg.st = make_vec2f(
-        ((atan2(intersection.x, intersection.z) / PI) + 1.f) * .5f,
-        -(asin(intersection.y) / PI) + .5f);
+    dg.st =
+        make_vec2f(((atan2(-intersection.x, intersection.z) / PI) + 1.f) * .5f,
+                   -(asin(intersection.y) / PI) + .5f);
```
hernando: I will trust you on this, but shouldn't there be a small test?

tribal-tec: Will have a follow-up commit for this. Testing the environment map in general as well.

brayns/tasks/AddModelFromBlobTask.cpp
tribal-tec: For more complicated loaders like for circuits, nesting of callbacks can or should be used today to mimic what you describe. The fixed amount for receiving the mesh can be tweaked for sure to account more the mesh loading/parsing itself, though if your link speed is slow, the upload takes more time. So it's not that easy to reflect that perfectly on the progress bar.

PR #665, opened 2019-01-22
Refactored simulation model to secondary model

favreau:
Geometry using the SECONDARY_MODEL_MATERIAL_ID are added to the secondary model.


engines/ospray/OSPRayModel.cpp
```diff
@@ -182,10 +183,17 @@ void OSPRayModel::_commitSpheres(const size_t materialId)
 
     ospCommit(geometry);
 
-    if (materialId == BOUNDINGBOX_MATERIAL_ID)
+    switch (materialId)
```
tribal-tec: this block shall be a function as it repeats some times

tribal-tec:
commit & PR message are wrong?

hernando:
I find this commit a bit strange, can you give more background on what are you trying to achieve? Maybe there are other ways

hernando:
The commit message is also incorrect because you're not refactoring anything, you're adding stuff as the simulation model was gone.

PR #667, opened 2019-01-22
Added properties to materials

favreau:
Adding properties to materials allows definition of extra properties that can be processed by the renderer.

### DIFF
--- a/brayns/engine/Model.h
+++ b/brayns/engine/Model.h
@@ -212,11 +212,10 @@ public:
     //@{
     virtual void commitGeometry() = 0;
 
-    /** Create a material with an assigned id and name.
-     *  No material properties are set.
-     */
-    BRAYNS_API MaterialPtr createMaterial(const size_t materialId,
-                                          const std::string& name);
+    /** Factory method to create an engine-specific material. */
+    BRAYNS_API virtual MaterialPtr createMaterial(
+        const size_t materialId, const std::string& name,
+        const PropertyMap& properties = {}) = 0;
 
     /**
      * Create a volume with the given dimensions, voxel spacing and data type
@@ -383,6 +382,13 @@ public:
        transparency, reflection, and light emission
     */
     void BRAYNS_API setMaterialsColorMap(const MaterialsColorMap colorMap);
+
+    /**
+     * @brief createMissingMaterials Checks that all materials exist for
+     * existing geometry in the model. Missing materials are created with the
+     * default parameters
+     */
+    BRAYNS_API void createMissingMaterials(const PropertyMap& properties = {});
 
     /**
      * @brief getMaterials Returns a reference to the map of materials handled
--- a/brayns/engine/Model.h
+++ b/brayns/engine/Model.h
@@ -213,9 +213,9 @@ public:
     virtual void commitGeometry() = 0;
 
     /** Factory method to create an engine-specific material. */
-    BRAYNS_API virtual MaterialPtr createMaterial(
-        const size_t materialId, const std::string& name,
-        const PropertyMap& properties = {}) = 0;
+    BRAYNS_API MaterialPtr createMaterial(const size_t materialId,
+                                          const std::string& name,
+                                          const PropertyMap& properties = {});
 
     /**
      * Create a volume with the given dimensions, voxel spacing and data type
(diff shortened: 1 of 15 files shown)


brayns/engine/Model.h
```diff
@@ -387,7 +388,7 @@ class Model
      * existing geometry in the model. Missing materials are created with the
      * default parameters
      */
-    BRAYNS_API void createMissingMaterials();
+    BRAYNS_API void createMissingMaterials(const PropertyMap& properties = {});
```
tribal-tec: that is not called other than this anyways, so the parameter is not needed for this function?!

favreau: Properties are forwarded to the material itself (line 558). It's true that Brayns never calls this method with properties, but this can still be done by a plugin for instance (that's actually what I do and why I need this extra parameter).

karjonas:
lgtm

engines/ospray/OSPRayModel.cpp
```diff
@@ -565,10 +565,12 @@ void OSPRayModel::commitMaterials(const std::string& renderer)
 }
 
 MaterialPtr OSPRayModel::createMaterial(const size_t materialId,
-                                        const std::string& name)
+                                        const std::string& name,
+                                        const PropertyMap& properties)
 {
     MaterialPtr material = std::make_shared<OSPRayMaterial>();
     material->setName(name);
+    material->setProperties(properties);
```
hernando: Add a constructor with these parameters parameters?

brayns/engine/Model.h
hernando: Shouldn't you pass the material type name as well?

tribal-tec: As discussed offline, this `createMissingMaterials()` shall be gone completely. The CircuitViewer plugin uses it though, but imo that shouldn't be necessary.  @hernando could you have a look there?

hernando: I'll do.

plugins/CircuitViewer/io/utils.h
```diff
@@ -27,8 +27,8 @@ namespace brayns
 {
 namespace
 {
-brion::GIDSet _gidsFromRange(const uint32_t first, const uint32_t last,
-                             const double fraction, const int32_t seed)
+inline brion::GIDSet _gidsFromRange(const uint32_t first, const uint32_t last,
```
favreau: I somewhat had to make this inline in order to avoid a 'defined but not used [-Werror=unused-function]' compilation error :-/

plugins/CircuitViewer/io/utils.h
```diff
@@ -59,8 +59,8 @@ brion::GIDSet _gidsFromRange(const uint32_t first, const uint32_t last,
     return brion::GIDSet(gids.begin(), gids.end());
 }
 
-brion::GIDSet _keyToGIDorRange(const std::string& key, const double fraction,
-                               const int32_t seed)
+inline brion::GIDSet _keyToGIDorRange(const std::string& key,
```
favreau: dito

brayns/engine/Scene.cpp
```diff
@@ -440,6 +440,9 @@ void Scene::_updateEnvironmentMap()
     }
 
     if (_backgroundMaterial->isModified())
+    {
+        _backgroundMaterial->resetModified(); // TODO????
```
tribal-tec: no, to my knowledge that's wrong. All the materials are reset at some point

plugins/CircuitViewer/io/MorphologyLoader.cpp
```diff
@@ -702,4 +704,32 @@ PropertyMap MorphologyLoader::getCLIProperties()
     pm.setProperty(PROP_USE_SDF_GEOMETRIES);
     return pm;
 }
+
+void createMissingMaterials(Model& model)
```
tribal-tec: that's still nuclear bomb to kill a fly :) The materials that are required to be created should be known in here, but @hernando said he will look into this.

brayns/engine/Scene.cpp
favreau: oops... removed

plugins/CircuitViewer/io/MorphologyLoader.cpp
favreau: I would make it a different PR.

hernando: Let me figure it out, I was the last one touching that code.

brayns/engine/Model.h
hernando: Done

plugins/CircuitViewer/io/MorphologyLoader.h
```diff
@@ -107,6 +107,8 @@ class MorphologyLoader : public Loader
 
     PropertyMap _defaults; // command line defaults
 };
+
+void createMissingMaterials(Model& model);
```
tribal-tec: rm this line aka revert this file

tribal-tec: still there?

Issue #588, opened 2018-10-11
If all spheres centers are coplanar in an XYZ file, nothing gets rendererd.


hernando:
A simple example:

     0 0 0
     1 0 0
     0 1 0
     1 1 0

Additionally, files without an end of line in the last line or empty lines are not accepted.

hernando:
@tribal-tec What about fixing this one for the next release? Seems simple.

tribal-tec:
I don't know if it's simple, but feel free to fix it then :smile:

PR #674, opened 2019-01-24
Fix for #588

hernando:
The problem was that the formula used to compute the sphere radius gave 0
when the bounding box volume was 0.
For those cases, a similar formula based on the bounding box area is used
instead.

### DIFF
--- a/brayns/io/XYZBLoader.cpp
+++ b/brayns/io/XYZBLoader.cpp
@@ -29,14 +29,19 @@
 
 #include <boost/filesystem.hpp>
 
-namespace
-{
-const auto LOADER_NAME = "xyzb";
-}
-
 namespace brayns
 {
-constexpr float DEFAULT_POINT_SIZE = 0.0001f;
+namespace
+{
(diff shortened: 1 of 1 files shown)


brayns/io/XYZBLoader.cpp
```diff
@@ -110,8 +116,12 @@ ModelDescriptorPtr XYZBLoader::importFromBlob(
     // Find an appropriate mean radius to avoid overlaps of the spheres, see
     // https://en.wikipedia.org/wiki/Wigner%E2%80%93Seitz_radius
     const auto volume = bbox.getSize().product();
-    const double meanRadius =
-        std::pow((3. / (4. * M_PI * (numlines / volume))), 1. / 3.);
+    const auto density4PI = 4 * M_PI * numlines /
+                            (volume > 1e-15 ? volume : _computeHalfArea(bbox));
```
tribal-tec: make constant for `1e-15` like `ALMOST_ZERO`?

PR #678, opened 2019-01-25
Fixed secondary model initialization and utilization

favreau:


engines/ospray/OSPRayRenderer.cpp
```diff
@@ -83,6 +83,8 @@ void OSPRayRenderer::commit()
         {
             auto& model =
                 static_cast<OSPRayModel&>(simulationModel->getModel());
+            ospSetObject(_renderer, "secondaryModel",
```
tribal-tec: move it below where the "world" is set

favreau: hum... it's specific to simulation...

tribal-tec: for your plugin maybe, but generically the second model could be used anything

PR #680, opened 2019-01-28
Fix loading of NPOT textures, support loading raw volumes from blob

tribal-tec:


brayns/common/ImageManager.cpp
```diff
@@ -32,36 +32,36 @@ Texture2DPtr ImageManager::importTextureFromFile(
...
     freeimage::SwapRedBlue32(image.get());
 #endif
 
-    Texture2DPtr texture(new Texture2D);
+    const auto width = FreeImage_GetWidth(image.get());
+    const auto height = FreeImage_GetHeight(image.get());
+    const auto pitch = FreeImage_GetPitch(image.get());
+    const auto channels = 4;
+
+    std::vector<unsigned char> rawData(height * pitch);
```
hernando: Is this correct? Shouldn't it be `height * pitch * channels`?

tribal-tec: Yes it is correct

brayns/engine/SharedDataVolume.cpp
```diff
@@ -78,9 +78,15 @@ void SharedDataVolume::mapData(const std::string& filename)
...
-void SharedDataVolume::mapData(const std::vector<char>& buffer)
+void SharedDataVolume::mapData(const uint8_ts& buffer)
 {
     _memoryBuffer.insert(_memoryBuffer.begin(), buffer.begin(), buffer.end());
     setVoxels(_memoryBuffer.data());
 }
+
+void SharedDataVolume::mapData(uint8_ts&& buffer)
+{
+    _memoryBuffer = std::move(buffer);
```
karjonas: Do you need the move here? Is this not already an rvalue reference?

brayns/common/material/Texture2D.cpp
```diff
@@ -35,4 +35,9 @@ void Texture2D::setRawData(unsigned char* data, size_t size)
     _rawData.clear();
     _rawData.assign(data, data + size);
 }
+
+void Texture2D::setRawData(std::vector<unsigned char>&& rawData)
+{
+    _rawData = std::move(rawData);
```
karjonas: Already moved, same as below?

brayns/engine/SharedDataVolume.cpp
hernando: Yes you need it, buffer is an rvalue reference in the declaration, but once inside the function body it becomes an lvalue because it's named and it can appear at the left side of an assignment.

brayns/common/ImageManager.cpp
hernando: In fact I was not talking about the bytes per Pixel, but the number of channels. But if you tell me that pitch is in bytes and not number of elements, that would explain why this code works (and would always do regardless of channels and bytes per pixel).

tribal-tec: That's what apparently the pitch in FreeImage is, yes.

PR #689, opened 2019-01-31
fix(js): ignore app bar height when calculating default snapshot height

rolandjitsu:


js/apps/viewer/src/app/snapshot/snapshot.tsx
```diff
@@ -404,13 +405,13 @@ function setMimeType(data: string, format?: ImageFormat) {
...
 
-function defaultSize(): {
+function defaultSize(currentWidth: Breakpoint): {
     width: number;
     height: number;
 } {
     const [width, height] = [
         window.innerWidth,
-        window.innerHeight
+        window.innerHeight - (isWidthDown('xs', currentWidth) ? 64 : 56)
```
tribal-tec: magic number looks wrong. why not using the current viewport of Brayns?

rolandjitsu: I made some consts now. We cannot use the viewport from Brayns because other clients could change that and the snapshot default size should be the size of the browser's viewport that the snapshot is taken in.

PR #692, opened 2019-01-31
Make loaders available during plugin initialization

karjonas:


brayns/Brayns.cpp
```diff
@@ -78,11 +78,10 @@ struct Brayns::Impl : public PluginAPI
         _registerKeyboardShortcuts();
         _setupCameraManipulator(CameraMode::inspect, false);
 
-        _pluginManager.initPlugins(this);
-
         _createFrameBuffer();
 
         _registerLoaders();
```
tribal-tec: we need to do the framebuffer stuff after plugins because of the opendeck framebuffer creation

karjonas: Done

PR #693, opened 2019-01-31
Fix snapshots with simulation data

tribal-tec:
* clone of models is needed for snapshot scene to have an immutable simulation
  handler and transfer function; the geometry though is still referenced to save
  memory
* animation time provided to snapshot RPC will load the correct frame,
  so changing the global animation is no longer needed (plus this does not
  guarantee to have the desired frame loaded for the snapshot)


plugins/CircuitViewer/tests/braynsTestData.cpp
```diff
@@ -113,11 +117,51 @@ BOOST_AUTO_TEST_CASE(circuit_with_simulation_mapping)
...
+            .getAnimationParameters());
+    params.animParams->setFrame(42);
+
+    auto image = clientServer.makeRequest<
+        brayns::SnapshotParams, brayns::ImageGenerator::ImageBase64>("snapshot",
+                                                                     params);
+    auto decodedImage = base64_decode(image.data);
+    const auto referenceFilename = BRAYNS_TESTDATA_IMAGES_PATH
+        "testdataallmini50basicsimulation_snapshot.png";
+#ifdef GENERATE_REFERENCE_SNAPSHOT
```
karjonas: There is already a define for generating new images?

plugins/CircuitViewer/tests/braynsTestData.cpp
```diff
@@ -36,7 +37,13 @@
...
 
 #define BOOST_TEST_MODULE braynsTestData
-#include <boost/test/unit_test.hpp>
+#include "tests/ClientServer.h"
+
+#include <boost/filesystem.hpp>
+namespace fs = boost::filesystem;
+#include <fstream>
+
+//#define GENERATE_REFERENCE_SNAPSHOT
```
karjonas: Remove?

plugins/Rockets/SnapshotTask.h
```diff
@@ -88,18 +105,17 @@ class SnapshotFunctor : public TaskFunctor
         else
             *_camera = engine.getCamera();
 
-        *_scene = engine.getScene();
+        _scene->cloneFrom(engine.getScene());
     }
 
     ImageGenerator::ImageBase64 operator()()
     {
+        _scene->commit();
```
karjonas: Any reason for this reordering?

plugins/CircuitViewer/tests/braynsTestData.cpp
```diff
@@ -113,11 +117,51 @@ BOOST_AUTO_TEST_CASE(circuit_with_simulation_mapping)
...
+    {
+        std::fstream file(newFilename, std::ios::out);
+        file << decodedImage;
+    }
+    const auto newImage{pdiff::read_from_file(newFilename)};
+    const auto referenceImage{pdiff::read_from_file(referenceFilename)};
+    const auto liveImage{pdiff::read_from_file(
+        BRAYNS_TESTDATA_IMAGES_PATH "testdataallmini50basicsimulation.png")};
+    BOOST_CHECK(pdiff::yee_compare(*referenceImage, *newImage));
+    BOOST_CHECK(!pdiff::yee_compare(*liveImage, *newImage));
```
hernando: Why don't you try to refactor some of this code to a compareTestImage function that could share code with the function in PDiffHelpers.h?

brayns/engine/Scene.cpp
```diff
@@ -63,37 +63,44 @@ std::shared_ptr<T> _remove(std::vector<std::shared_ptr<T>>& list,
...
+
+        _modelDescriptors.clear();
+        _modelDescriptors.reserve(rhs._modelDescriptors.size());
+        for (const auto& modelDesc : rhs._modelDescriptors)
+        {
+            auto newModelDesc =
+                std::make_shared<ModelDescriptor>(createModel(),
+                                                  modelDesc->getPath());
+            newModelDesc->cloneFrom(*modelDesc);
+            _modelDescriptors.push_back(newModelDesc);
```
hernando: Can't you simplify lines 88:92 to `_modelDescriptors.push_back(modelDesc->clone())`?

plugins/Rockets/SnapshotTask.h
tribal-tec: To comply what the 'normal' brayns rendering loop does, otherwise no real reason that would fix anything.

brayns/engine/Model.h
```diff
@@ -443,43 +451,61 @@ class Model
...
     bool _trianglesMeshesDirty{true};
-    Boxd _trianglesMeshesBounds;
-
-    StreamlinesDataMap _streamlines;
     bool _streamlinesDirty{true};
-    Boxd _streamlinesBounds;
-
-    Boxd _bounds;
+    bool _sdfGeometriesDirty{true};
+    bool _volumesDirty{true};
```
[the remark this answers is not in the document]
tribal-tec: Then the original model would have to re-commit all its geometries as well for no reason if the dirty flags are shared.

plugins/Rockets/SnapshotTask.h
```diff
@@ -88,18 +105,17 @@ class SnapshotFunctor : public TaskFunctor
         else
             *_camera = engine.getCamera();
 
-        *_scene = engine.getScene();
+        _scene->cloneFrom(engine.getScene());
```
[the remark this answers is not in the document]
tribal-tec: I reused the `cloneFrom` term from the property maps. The virtual clone is worse imo rather than dealing with the cloning in the base class.

brayns/engine/Model.h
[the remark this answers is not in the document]
hernando: I see, but that in fact implies that commit should be thread safe.
Do you plan to do copy on write if the reference count is not one at the moment the geometry is modified?

plugins/CircuitViewer/tests/braynsTestData.cpp
tribal-tec: Because that code was first (see snapshot.cpp), so once someone did compareTestImage did not unify in the first place.

hernando: Are you implying that it should be @karjonas who does it?

hernando: I see, I didn't understand it that way. It's true that if you take the code from snapshot, you're seemingly doing the right thing.

But are you going to give it a try or leave it for another PR?

tribal-tec: I can try, yes.

brayns/engine/Scene.cpp
tribal-tec: Probably I can.

plugins/Rockets/SnapshotTask.h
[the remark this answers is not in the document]
tribal-tec: The creation of scenes is part of the engine, so there is no way to create a scene from within a scene clone() function. Changing this should be done for all engine entities then, not just the scene.

brayns/engine/Model.h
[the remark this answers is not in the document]
tribal-tec: No plan there, at most modifying the geometries while they are shared should be forbidden. But not in this PR, was not safe until now as well and technically not possible for web clients.

plugins/Rockets/SnapshotTask.h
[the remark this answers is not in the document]
hernando: Fair enough

tribal-tec:
Good to go?

karjonas:
lgtm, if @hernando approves.

PR #695, opened 2019-02-01
Fix radius property having no effect for point-cloud models

tribal-tec:


tribal-tec:
Yes, that was the issue.

PR #696, opened 2019-02-04
Fix Rockets deadlock in plugins

karjonas:
A deadlock can appear when doing scene modifications with callbacks to rockets in a plugin method that was initially called from rockets. Fixed by adding the proper scope guards.

### DIFF
--- a/plugins/Rockets/RocketsPlugin.cpp
+++ b/plugins/Rockets/RocketsPlugin.cpp
@@ -264,7 +264,9 @@ public:
                               const PropertyMap& input,
                               const std::function<void(PropertyMap)>& action)
     {
-        _jsonrpcServer->connect(desc.methodName, [action](const auto& request) {
+        _jsonrpcServer->connect(desc.methodName, [action,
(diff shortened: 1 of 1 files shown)


plugins/Rockets/RocketsPlugin.cpp
```diff
@@ -284,8 +290,9 @@ class RocketsPlugin::Impl : public ActionInterface
                          const PropertyMap& input, const PropertyMap& output,
                          const std::function<PropertyMap(PropertyMap)>& action)
     {
-        _bindEndpoint(desc.methodName, [ name = desc.methodName,
-                                         action ](const auto& request) {
+        _bindEndpoint(desc.methodName, [ name = desc.methodName, action,
+                                         this ](const auto& request) {
+            ScopedCurrentClient scope(this->_currentClientID, request.clientID);
```
tribal-tec: not needed imo, as `_bindEndpoint` does it.

karjonas: Indeed

PR #698, opened 2019-02-06
refactor(js): refactor state management for some of the components

rolandjitsu:


js/apps/viewer/src/app/scene/scene.tsx
```diff
@@ -498,6 +498,7 @@ export class SceneModels extends PureComponent<Props, State> {
                         return {clipPlanes};
                     });
                 }),
+            // TODO: Remove this one!
```
tribal-tec: TODO still valid?

rolandjitsu: Yes, I wasn't sure how to handle that right now.

PR #700, opened 2019-02-07
Fixed lights issue in snapshots

favreau:


favreau:
Also fixes the fact that lights updates are always 1 frame late in the default rendering loop.

PR #705, opened 2019-02-07
Use glm instead of vmmlib

tribal-tec:


brayns/common/mathTypes.h
```diff
@@ -21,47 +21,147 @@
...
+        return _min == other._min && _max == other._max;
+    }
+
+    void merge(const Box<T>& aabb)
+    {
+        const auto& min = aabb.getMin();
+        const auto& max = aabb.getMax();
+
+        if (min.x < _min.x)
+            _min.x = min.x;
```
karjonas: I prefer _min.x = std::min(_min.x, min.x); instead of if cases.

favreau:
💯 👍

karjonas:
:100 :1st_place_medal:

PR #709, opened 2019-02-11
fix(js): do not use ResizeObserver to observe element rect

rolandjitsu:


rolandjitsu:
@favreau our CIs are down, so to get the fix for the TF just use my fork locally.

favreau:
@rolandjitsu Super cool. Many thanks for the quick fix.

PR #711, opened 2019-02-11
Improve AnimationParameters

tribal-tec:
* frameCount instead of start/end
* playing state to support different delta values from the frontend


tribal-tec:
Retest this please

PR #717, opened 2019-02-14
Workaround for https://github.com/assimp/assimp/issues/2337

tribal-tec:


tribal-tec:
Yes, the version before the broken commit.
Downloading things during cmake phase is not something that works with Nix, at least it's not recommended iirc.

hernando:
It's true that Adrien commented bitterly against it, but I don't remember the exact reason. In any case, for testing your code you shouldn't  need such a big mesh (even it it's the mesh that breaks assimp). Following that logic you should add tests for all possible bugs/regressions in third party dependencies.

PR #725, opened 2019-02-19
Fix load order dependency problem with OpenDeck and VRPN plugin

tribal-tec:


plugins/OpenDeck/OpenDeckPlugin.cpp
```diff
@@ -47,27 +98,26 @@ OpenDeckPlugin::OpenDeckPlugin(const Vector2ui& wallRes,
 
 void OpenDeckPlugin::init()
 {
+    auto& engine = _api->getEngine();
 #ifdef BRAYNS_USE_OSPRAY
-    _api->getEngine().addCameraType("cylindric");
-    _api->getEngine().addCameraType("cylindricStereo");
-    _api->getEngine().addCameraType("cylindricStereoTracked");
+    engine.addCameraType("cylindric");
+    engine.addCameraType("cylindricStereo", getCylindricStereoProperties());
```
hernando: The properties for this camera were missing before, weren't they?

tribal-tec: No, they were added by the VRPN plugin. But for this work, the VPRN had to be loaded after the OpenDeck one.

tribal-tec: Ah, for the non-tracked camera, probably they weren't added, yes.

PR #734, opened 2019-02-26
Release preparation

tribal-tec:


rolandjitsu:
Don't forget to also bump the js stuff, I think you can run `yarn release 0.8.0` from the root of the js folder, but I'm not sure 😄

PR #735, opened 2019-02-26
Make sure a gcc version compatible with Optix can be found

hernando:


engines/optix/CMakeLists.txt
```diff
@@ -8,6 +8,11 @@
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/CMake)
 
 common_find_package(CUDA REQUIRED SYSTEM)
+if(CUDA_FOUND)
+  # This doesn't from common_find_package_post unless we export some variables
```
karjonas: This doesn't what?

PR #739, opened 2019-03-06
bug(optix): do not memcpy from empty source

karjonas:


engines/optix/OptiXModel.cpp
```diff
@@ -39,9 +39,11 @@ void setBufferRaw(RTbuffertype bufferType, RTformat bufferFormat,
     auto context = OptiXContext::get().getOptixContext();
     if (!buffer)
         buffer = context->createBuffer(bufferType, bufferFormat, bufferSize);
-
-    memcpy(buffer->map(), src, bufferSize);
-    buffer->unmap();
+    if (src != nullptr && bufferSize > 0)
```
tribal-tec: shouldn't the `bufferSize > 0` check also go in the previous if? maybe a combined if all conditions are true, then do something?

karjonas: I think creating a zero-sized buffer is preferred over creating no buffer at all since OptiX usually fails if buffers are unset even if they are unused.

karjonas:
retest this please

PR #740, opened 2019-03-08
bug(optix): fix snapshot crash in OptiX

karjonas:
OptiX only supports one context per process so instead of cloning the scene and camera, the OptiX screenshots are rendered serially and reuses the same context.

### DIFF
--- a/apps/ui/Application.cpp
+++ b/apps/ui/Application.cpp
@@ -388,6 +388,8 @@ void Application::render()
     glClearColor(0.f, 0.f, 0.f, 1.f);
     glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
 
+    auto lock{m_brayns.getEngine().getRenderScopeLock()};
+
     size_t offset = 0;
     size_t bufferIdx = 0;
     for (auto frameBuffer : m_brayns.getEngine().getFrameBuffers())
--- a/plugins/Rockets/SnapshotTask.cpp
+++ b/plugins/Rockets/SnapshotTask.cpp
@@ -1,0 +1,188 @@
+/* Copyright (c) 2015-2019, EPFL/Blue Brain Project
+ * All rights reserved. Do not distribute without permission.
+ *
+ * This file is part of Brayns <https://github.com/BlueBrain/Brayns>
+ *
+ * This library is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU Lesser General Public License version 3.0 as published
+ * by the Free Software Foundation.
+ *
+ * This library is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include <brayns/common/tasks/Task.h>
+
+#include "ImageGenerator.h"
+#include <brayns/common/utils/utils.h>
+#include <brayns/engine/Camera.h>
+#include <brayns/engine/Engine.h>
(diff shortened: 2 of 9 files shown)


plugins/Rockets/SnapshotTask.cpp
```diff
@@ -0,0 +1,188 @@
...
+        _camera = engine.getCameraPtr();
+        _renderer = engine.getRendererPtr();
+    }
+}
+
+ImageGenerator::ImageBase64 SnapshotFunctor::_renderSnapshot(
+    const size_t numAccumulations)
+{
+    const auto nameShort =
+        _params.name.empty() ? "" : " " + shortenString(_params.name);
```
hernando: Type deduction rules must be really complicated if this can successfully infer that nameShort is std::string...

plugins/Rockets/SnapshotTask.cpp
```diff
@@ -0,0 +1,188 @@
...
+{
+    _scene->commit();
+
+    _camera->updateProperty("aspect", double(_params.size.x) / _params.size.y);
+    _camera->commit();
+
+    if (_params.renderingParams)
+    {
+        _params.renderingParams->setSamplesPerPixel(1);
+        _params.renderingParams->setSubsampling(1);
```
hernando: Doesn't it work using samples per pixels for OSPRay as well?

plugins/Rockets/SnapshotTask.cpp
```diff
@@ -0,0 +1,188 @@
...
+    _renderer->setCamera(_camera);
+    _renderer->setScene(_scene);
+    _renderer->commit();
+
+    return _renderSnapshot(_params.samplesPerPixel);
+}
+
+ImageGenerator::ImageBase64 SnapshotFunctor::_takeSnapshotSerial()
+{
+    auto lock{_engine.getRenderScopeLock()};
```
hernando: I don't see where you are setting the snapshot size in this case. In the concurrent snapshot the rendering parameters (with the buffer size) are passed to the constructor of the new renderer, but here you use the same renderer.

apps/ui/Application.cpp
```diff
@@ -337,6 +337,8 @@ void Application::render()
 
     m_brayns.commitAndRender();
 
+    auto lock{m_brayns.getEngine().getRenderScopeLock()};
```
hernando: This looks like a hack that will reduce maintainability. I leave it up to @tribal-tec to decide whether making the interface more function programming oriented would be better or not, but I would try to find another solution.

plugins/Rockets/SnapshotTask.cpp
karjonas: It does but OSPRay uses accumulation frames instead to get a nice progress bar.

plugins/Rockets/SnapshotTask.cpp
karjonas: It is set on line 112 when creating the buffer.

plugins/Rockets/SnapshotTask.h
```diff
@@ -56,121 +56,20 @@ class SnapshotFunctor : public TaskFunctor
...
+
     SnapshotParams _params;
-    FrameBufferPtr _frameBuffer;
-    CameraPtr _camera;
-    RendererPtr _renderer;
-    ScenePtr _scene;
+    FrameBufferPtr _frameBuffer{nullptr};
+    CameraPtr _camera{nullptr};
+    RendererPtr _renderer{nullptr};
+    ScenePtr _scene{nullptr};
```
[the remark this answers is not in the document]
karjonas: Done

apps/ui/Application.cpp
hernando: Any comments here @tribal-tec?

tribal-tec: I don't see why the lock is needed where there is locking inside Brayns.cpp. But the whole change is suspicious to me, as suddenly an engine affects how things have to be done from the outside. I don't like in general how Brayns has to change to work with Optix. Ideally, all the hacks should be only within the engine implementation itself.

ppodhajski:
retest this please

ppodhajski:
retest this please

ppodhajski:
retest this please

ppodhajski:
retest this please

ppodhajski:
retest this please

PR #733, opened 2019-02-22
Ospray 1.8

hernando:

PR #744, opened 2019-03-12
bug(ospray): do not commit empty vectors

karjonas:
When running OSPRay in debug mode this triggers an assertion failure


hernando:
This if fixed by #733 with a better approach, it will try to get CI passing today for merging it.

PR #745, opened 2019-03-14
refactor(ospray): add helper methods for setting OSPRay properties

karjonas:


engines/ospray/OSPRayMaterial.cpp
```diff
@@ -136,10 +134,10 @@ OSPTexture OSPRayMaterial::_createOSPTexture2D(Texture2DPtr texture)
 
     OSPTexture ospTexture = ospNewTexture("texture2d");
 
-    const osp::vec2i size{int(texture->getWidth()), int(texture->getHeight())};
+    const Vector2i size{int(texture->getWidth()), int(texture->getHeight())};
 
-    ospSet1i(ospTexture, "type", static_cast<int>(type));
-    ospSet2i(ospTexture, "size", size.x, size.y);
+    osphelper::set(ospTexture, "type", static_cast<int>(type));
+    osphelper::set(ospTexture, "size", size);
```
hernando: Could this work:

    osphelper::set(ospTexture, "size", {int(texture->getWidth()), int(texture->getHeight())});

engines/ospray/utils.cpp
```diff
@@ -193,4 +194,76 @@ void addInstance(OSPModel rootModel, OSPModel modelToAdd,
...
+{
+    ospSet1i(obj, id, v);
+}
+void set(OSPObject obj, const char* id, size_t v)
+{
+    ospSet1i(obj, id, v);
+}
+void set(OSPObject obj, const char* id, const Vector2f& v)
+{
+    ospSet2f(obj, id, v.x, v.y);
```
hernando: Is it faster/better to use `ospSet2fv(obj, id, glm::value_ptr(v));`

engines/ospray/utils.cpp
```diff
@@ -193,4 +194,76 @@ void addInstance(OSPModel rootModel, OSPModel modelToAdd,
...
+{
+    ospSetString(obj, id, s.c_str());
+}
+void set(OSPObject obj, const char* id, float v)
+{
+    ospSet1f(obj, id, v);
+}
+void set(OSPObject obj, const char* id, double v)
+{
+    ospSet1f(obj, id, v);
```
hernando: I wouldn't make these demoting casts implicit.

engines/ospray/utils.h
```diff
@@ -63,4 +63,30 @@ std::vector<std::array<float, S>> convertVectorToFloat(
     }
     return output;
 }
-}
+
+namespace osphelper
+{
```
hernando: I think it's better to make these functions inline. It wouldn't be a premature optimization, it's about avoiding unreasonable penalizations.

engines/ospray/OSPRayMaterial.cpp
karjonas: No

engines/ospray/utils.cpp
karjonas: yes and yes

engines/ospray/utils.cpp
karjonas: done

engines/ospray/utils.h
karjonas: I prefer to keep them in the cpp file to keep the header clean and since it is not performance critical I think that trumps the performance loss.

karjonas:
Issues fixed

PR #748, opened 2019-03-20
Add the multiview camera plugin

chevtche:


plugins/Multiview/MultiviewPlugin.cpp
```diff
@@ -0,0 +1,70 @@
...
+    {
+        throw std::runtime_error(
+            "The multiview camera arm length must be stricly positive");
+    }
+}
+
+void MultiviewPlugin::init()
+{
+    auto& engine = _api->getEngine();
+#ifdef BRAYNS_USE_OSPRAY
```
karjonas: I think you should look at the runtime engine instead, like ```if (params.getApplicationParameters().getEngine() == "ospray")```

karjonas:
All copyrights should be 2019. I also think the responsible author is pointless.

PR #749, opened 2019-03-21
feat(cpp): add light manager

karjonas:
* Add a light manager to the scene
* Add spotlight and quadlight light sources, only supported in OSPRay for now
* Expose adding/removing of lights as RPCs


brayns/common/light/Light.h
```diff
@@ -18,40 +17,43 @@
...
-    BRAYNS_API void setIntensity(const float intensity)
-    {
-        _intensity = intensity;
-    }
-    BRAYNS_API float getIntensity() const { return _intensity; }
-protected:
-    BRAYNS_API Light(const Vector3f& color, float intensity);
+    Light() = default;
+    Light(const LightType type, const Vector3f& color, const float intensity,
+          const int id);
```
hernando: Without considering how the RPC interface would look for the moment, is it necessary to make the ID part of the light? Shouldn't `LightManager::addLight` return the ID instead?
That seems more consistent with the model API in my opinion. Models don't contain their own ID, it belong to `ModelDescriptor` and  which is returned by `Scene::addModel`.

brayns/engine/LightManager.h
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * @brief addDirectionalLight Creates a light that shines from a specific
+     * direction not from a specific position. This light will behave as though
+     * it is infinitely far away and the rays produced from it are all parallel
+     * @param direction Light source direction
+     * @param color Light source RGB color
+     * @param intensity Amount of light emitted
+     */
+    LightPtr addDirectionalLight(const Vector3f& direction,
```
hernando: This `addXYZLight` should be either free functions or static member functions named `createXYZLight`. I would even move them to the Light class. For adding the light, client code would only use `LightManager::addLight`

brayns/engine/LightManager.h
```diff
@@ -0,0 +1,131 @@
...
+                                 const Vector3f& color, float intensity);
+
+    /**
+     * @brief addSphereLight Createas a spehre light. The sphere light (or the
+     * special case point light) is a light emitting uniformly in all
```
hernando:
```suggestion
     * special case point light) is a positional light emitting uniformly in all
```

brayns/engine/LightManager.h
```diff
@@ -0,0 +1,131 @@
...
+     * @param radius The size of the sphere light
+     * @param color Light source RGB color
+     * @param intensity Amount of light emitted
+     */
+    LightPtr addSphereLight(const Vector3f& position, float radius,
+                            const Vector3f& color, float intensity);
+
+    /**
+     * @brief addQuadLight Creates a quad light. The quad light is a planar,
+     * procedural area light source emitting uniformly on one side into the
```
hernando: procedural?

brayns/engine/LightManager.h
```diff
@@ -0,0 +1,131 @@
...
+     * emitting into a cone of directions.
+     * @param position The center of the spotlight, in world-space
+     * @param direction  Main emission direction of the spot
+     * @param openingAngle  Full opening angle (in degree) of the spot; outside
+     * of this cone is no illumination
+     * @param penumbraAngle Size (angle in degree) of the "penumbra", the region
+     * between the rim (of the illumination cone) and full intensity of the
+     * spot; should be smaller than half of 'openingAngle'
+     * @param radius The size of the spotlight, the radius of a disk with normal
+     * 'direction'
```
hernando: I don't fully understand why radius does. The position, axis direction and angle are sufficient to specify an infinite cone. What is the radius doing?

brayns/engine/LightManager.h
hernando: This is not entirely clear to me. If each point in the light surface emit uniformly in the half-space, the integral over the area doesn't produce a uniform distribution over a solid angle (as I understand the sphere light does). For that to happen the area needs to be infinite. I would be more careful with the wording.

brayns/engine/LightManager.h
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * @brief addQuadLight Creates a quad light. The quad light is a planar,
+     * procedural area light source emitting uniformly on one side into the
+     * half-space.
+     * @param position World-space position of one vertex of the quad light
+     * @param edge1 Vector to one adjacent vertex
+     * @param edge2 Vector to the other adjacent vertex
+     * @param color Light source RGB color
+     * @param radius The size of the sphere light
```
hernando: copy pasta

brayns/engine/LightManager.h
```diff
@@ -0,0 +1,131 @@
...
+     * @param color Light source RGB color
+     * @param intensity Amount of light emitted
+     */
+    LightPtr addSphereLight(const Vector3f& position, float radius,
+                            const Vector3f& color, float intensity);
+
+    /**
+     * @brief addQuadLight Creates a quad light. The quad light is a planar,
+     * procedural area light source emitting uniformly on one side into the
+     * half-space.
```
hernando: On which half-space is the light emitting? You could define this based on the direction of the cross product of edge1 and edge2 for example.

brayns/engine/LightManager.h
```diff
@@ -0,0 +1,131 @@
...
+    LightPtr addQuadLight(const Vector3f& position, const Vector3f& edge1,
+                          const Vector3f& edge2, const Vector3f& color,
+                          float intensity);
+
+    /**
+     * @brief addSpotLight Creates a spot light. The spotlight is a light
+     * emitting into a cone of directions.
+     * @param position The center of the spotlight, in world-space
+     * @param direction  Main emission direction of the spot
+     * @param openingAngle  Full opening angle (in degree) of the spot; outside
```
hernando: Is it better to use degrees than radians?

brayns/common/light/Light.h
hernando: Now considering the json serialization, it seems easier the way you did it, but I'm not super happy about the idea of having the ID in the light. Anyway, I would remove the id from the constructor, make LightManager a friend and the only class capable of setting the ID and make LightManager::addLight return the ID.

plugins/Rockets/jsonSerialization.h
```diff
@@ -79,7 +80,43 @@ struct ModelTransferFunction
     size_t id;
     TransferFunction transferFunction;
 };
-}
+
+struct RPCDirectionalLight
```
hernando: There aren't any static structures for the different camera types. Why do you want them for lights?

After continuing reading, I've realized that it must be for the RPC methods for adding lights, right? Check with @tribal-tec, because the same way these are not needed for changing the camera type and properties, they may neither be necessary for lights. The key could be in using PropertyObject instead of PropertyMap, but maybe there's no way to validate the properties in this case.

karjonas:
I rewrote this PR a bit so that each light is actually a different class without any property map. The light manager now stores the ids for the added lights. The lights serialization is done with a generic RPCLight class which is populated in the rockets plugin.

brayns/engine/LightManager.h
karjonas: I leave it like this to mirror OSPRay properties.

PR #751, opened 2019-03-22
Handle unexpected Deflect server disconnections gracefully

hernando:
This commit combined with 038e139 in Deflect master prevent an
abort within libuv when the socket fd gets closed by QTcpSocket
after a server side disconnect.

### DIFF
--- a/plugins/Deflect/DeflectPlugin.cpp
+++ b/plugins/Deflect/DeflectPlugin.cpp
@@ -175,6 +175,8 @@ private:
     void _setupSocketListener()
     {
 #ifdef BRAYNS_USE_LIBUV
+        assert(_stream->isConnected());
+
         auto loop = uvw::Loop::getDefault();
         _pollHandle = loop->resource<uvw::PollHandle>(_stream->getDescriptor());
 
@@ -184,6 +186,11 @@ private:
         });
 
         _pollHandle->start(uvw::PollHandle::Event::READABLE);
+
+        _stream->setDisconnectedCallback([&] {
+            _pollHandle->stop();
+            _pollHandle.reset();
+        });
 #endif
     }
 


plugins/Deflect/DeflectPlugin.cpp
```diff
@@ -184,6 +186,8 @@ class DeflectPlugin::Impl
         });
 
         _pollHandle->start(uvw::PollHandle::Event::READABLE);
+
+        _stream->setDisconnectedCallback([&] { _pollHandle->stop(); });
```
tribal-tec: capture `_pollHandle` by copy, so it's always safe to be valid

hernando: Since `_pollHandle` is a member varaible, and [&] captures `this` be reference, this should be safe as well, shouldn't it?

tribal-tec: Not if `_pollHandle` is destructed before `_stream`? Also `&` is capturing more than you need.

hernando: I tried [=, _pollHandle], but it doesn't compile. I'll try to get back to this after the weekend.

tribal-tec: just `[_pollHandle]` should work?!

tribal-tec: `[pollHandle=_pollHandle]` does work in fact

tribal-tec: while testing and looking at the code again, it probably should capture the pollhandle by reference and also `.reset()` it, so that `_closeStream()` does not stop the handle again. I at least faced some error during a reconnect phase:
```
[INFO ] Closing Deflect stream
braynsService: src/unix/core.c:896: uv__io_stop: Assertion `loop->watchers[w->fd] == w' failed.
Aborted
```

PR #761, opened 2019-04-25
Added axis and grid predefined scenes

favreau:


brayns/engine/Scene.cpp
```diff
@@ -257,13 +255,124 @@ void Scene::visitModels(const std::function<void(Model&)>& functor)
         functor(modelDescriptor->getModel());
 }
 
-void Scene::buildDefault()
+void Scene::buildGrid()
```
karjonas: Since we now have a few demo scenes maybe we should move these builders into a new file?

favreau:
Thanks for the comments. As discussed, this will be moved to a dedicated plugin.

PR #765, opened 2019-04-26
fix(cpp): pass correct buffer sizes for meshes in OptixEngine

tribal-tec:
Allows to load bigger meshes; ~30% less memory is being used.
Also consider 'compact' bvh flag from commandline to use 'Sbvh' from Optix.


tribal-tec:
retest this please

ppodhajski:
retest this please

PR #771, opened 2019-05-22
fix(cpp): Rename default material to match OSPRay

karjonas:
This fixes a bug where changing to a raycasting renderer caused a crash in OSPRay


karjonas:
retest this please

PR #778, opened 2019-06-06
feat(optix): Support different camera programs

karjonas:


ppodhajski:
retest this please

engines/optix/OptiXContext.h
```diff
@@ -24,12 +24,16 @@
...
 
+#include "OptiXCameraProgram.h"
+
 #include <memory>
 #include <mutex>
 #include <unordered_map>
 
 namespace brayns
 {
+class OptiXCamera;
```
tribal-tec: unused

engines/optix/OptiXContext.h
```diff
@@ -84,7 +91,8 @@ class OptiXContext
 
     ::optix::Context _optixContext{nullptr};
 
-    std::map<std::string, OptixShaderProgram> _rendererProgram;
+    std::map<std::string, std::shared_ptr<OptixShaderProgram>> _rendererProgram;
```
tribal-tec: used typedefs OptixShaderProgramPtr and OptixCameraProgramPtr

engines/optix/OptiXEngine.cpp
```diff
@@ -88,6 +89,10 @@ void OptiXEngine::_createCameras()
                            0.0635,
                            {"Eye separation"}};
 
+    OptiXContext& context = OptiXContext::get();
+
+    auto camera = std::make_shared<OptiXPerspectiveCamera>();
```
tribal-tec: well, don't we have only "perspective" camera here? why adding 3 more that use the same program?

ppodhajski:
retest this please

ppodhajski:
retest this please

ppodhajski:
retest this please

engines/optix/OptiXEngine.cpp
karjonas: Yes, they use the same program but each camera has different properties which might not make sense to expose in all camera types. Ideally each camera would use its own program instead of the uber-camera program, but that has to be implement first. So I think it makes sense to keep it this way.

tribal-tec: I don't see where the perspective camera cuda code does orthographic or panoramic projection. But your choice to keep it like this.

karjonas: Heh, you are right. I assumed the camera types actually worked but apparently not. I removed the other ones so we only have the perspective camera.

PR #779, opened 2019-06-11
fix normal discontinuity in SDF conePills

Speierers:
Cone pills implement a smooth profile (sigmoid-like). Unfortunately, this creates discontinuities in the normals as shown on this render:
current_blend

This pull request fixes this:
smoother

As explained in this [article](https://en.wikipedia.org/wiki/Smoothstep) we could use either the `smooth()` function, or the `smoother()` function, which have slightly different stepness.

Here using smooth:
smooth

I am not certain what is the better solution in this case.


ppodhajski:
retest this please

PR #797, opened 2019-07-19
CMake fixes

karjonas:
Always add BRAYNS_NETWORK_ENABLED option

Make networking depend on libjpegturbo


CMakeLists.txt
```diff
@@ -59,11 +59,22 @@ common_find_package(LibJpegTurbo)
...
   option(BRAYNS_NETWORKING_ENABLED "Activate networking interfaces" ON)
-  if(BRAYNS_NETWORKING_ENABLED)
+else()
+  option(BRAYNS_NETWORKING_ENABLED "Activate networking interfaces" OFF)
+endif()
+
+if(BRAYNS_NETWORKING_ENABLED)
+  # We require turbojpeg for networking to get image streaming working
+  common_find_package(LibJpegTurbo REQUIRED)
+  if(ROCKETS_FOUND AND ROCKETS_USE_LIBWEBSOCKETS)
```
tribal-tec: why not for consistency with other options do `common_find_package(Rockets REQUIRED)`? If it's enabled, it shall error imo, same if `ROCKETS_USE_LIBWEBSOCKETS` is not defined

karjonas: Done

PR #806, opened 2019-08-09
Remove Boost.Filesystem and Boost.Any

tribal-tec:


brayns/io/ArchiveLoader.cpp
```diff
@@ -170,18 +166,25 @@ void extractFile(const std::string& filename, const std::string& destination)
...
     auto archive = _openArchive(blob);
     if (!archive)
         throw std::runtime_error("Blob is not a supported archive type");
-    _extractArchive(archive, fs::basename(blob.name), destination);
+    _extractArchive(archive, fs::path(blob.name).stem().string(), destination);
 }
 
 const auto LOADER_NAME = "archive";
+
+struct TmpFolder
```
karjonas: I would write it like this to make it a bit clearer what is happening:

```
struct TmpFolder
{
    TmpFolder()
    {
        char pathTemplate[] = "/tmp/extracted_XXXXXX";
        path = mkdtemp(pathTemplate);
    }
    ~TmpFolder() { fs::remove_all(path); }
    std::string path;
};
```

karjonas: While we are at it we should change it to /tmp/brayns/extracted_XXXXXX to decrease the pollution.

tribal-tec: well, it's deleted after loading, so not too much pollution.

`mkdtemp` modifies the input `pathTemplate`, hence it has to stay member. The returned path is the one that is temporary and shall not be deleted by `std::string` for instance. So not sure if there's a different way to implemented that.

karjonas: It is minor but I think it is nice to have a brayns prefix to know what program is creating the folders.

The constructor should be fine since assigning a char* to a std::string will make a deep copy. So the temporary `pathTemplate` is copied to `path`. A check for `mkdtemp` returning nullptr might be good though :)

tribal-tec: Doesn't work. I keep the current working solution :smile:

karjonas: Okay then

tribal-tec:
I'm still testing the docker build and OSX, but the Apple one looks tough as their clang standard library has no filesystem...

PR #860, opened 2019-11-14
Various fixes

favreau:


favreau:
I only removed the duplicated code, not the feature.

PR #861, opened 2019-11-14
Add support for anterograde tracing

NadirRoGue:
- Added support for anterograde tracing (staining)
- Added set-material-range entry point (set a list of materials IDs to the same material)
- Removed unused colon in marco FROM_JSON

### DIFF
--- a/plugins/CircuitExplorer/plugin/CircuitExplorerPlugin.cpp
+++ b/plugins/CircuitExplorer/plugin/CircuitExplorerPlugin.cpp
@@ -64,6 +64,10 @@
 #define REGISTER_LOADER(LOADER, FUNC) \
     registry.registerLoader({std::bind(&LOADER::getSupportedDataTypes), FUNC});
 
+const std::string ANTEROGRADE_TYPE_AFFERENT = "afferent";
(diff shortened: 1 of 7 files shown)


plugins/CircuitExplorer/plugin/CircuitExplorerPlugin.cpp
```diff
@@ -405,6 +414,14 @@ void CircuitExplorerPlugin::init()
             "make-movie",
             [&](const MakeMovieParameters& params) { _makeMovie(params); });
 
+        PLUGIN_INFO << "Registering 'trace-aterograde' endpoint" << std::endl;
```
tolokoban: Small typo here: __trace-anterograde__ instead of __trace-aterograde__
`PLUGIN_INFO << "Registering 'trace-anterograde' endpoint" << std::endl;`

NadirRoGue: Fixed

NadirRoGue:
> It could be great if we can apply a same material to everything but just passing an empty list as **materialIds**.
> Like this, we could be able to display one cell in green and all other cells in grey, for example.

done

PR #863, opened 2019-11-26
Addons and fixes

NadirRoGue:
- Fixed set_material_range python method (wrong parameter name)
- Using atomic flags for controlled streaming method
- Added random colouring through set-material-range
- Added exit-later entry point (will shutdown Brayns after a specific amount of minutes, allows reschedule)


NadirRoGue:
retest this please

PR #864, opened 2019-12-18
Plugin additions and some fixes

NadirRoGue:
- Eased BlueConfig/CircuitConfig file naming restrictions for loading
- Added original circuit path to Brayns cache files (.brayns)
- Added optinal parameter to the snapshot function to save the result to a local file
- Added CircuitInfo plugin (queries circuit info such as reports, targets, cell GIDs, etc.)
- Added CircuitRenderer plugin (allows for independent rendereres to be developed and used with other plugins)


ppodhajski:
retest this please

PR #869, opened 2020-01-21
Remove one const in cast to compile with gcc8.

matz-e:


matz-e:
Rebased. I'm also not sure that I have write/merge permissions to this repository.

PR #873, opened 2020-04-17
Hot colorscheme switch

NadirRoGue:
- Added support to remap circuit color to a different scheme on the fly
- Meshes and circuit (for some schemes) metadata contains now material information
- OSPRay engine SDF rendering minor optimizations
- CMake fixes to import HighFive target when duplicated
- Fixed snapshot utility to generate images with simulation data
- Bumped Brion version to lastest (now with support for new sonata readers)
- Updated CMake/common submodule to fix recursive clonning


NadirRoGue:
retest this please

PR #913, opened 2021-04-29
Add entrypoint to get loaded plugins

Adrien4193:
Add plugins list to ApplicationParameters and expose them to the REST API.


plugins/Rockets/jsonSerialization.h
```diff
@@ -546,6 +546,7 @@ inline void init(brayns::Scene* s, ObjectHandler* h)
 inline void init(brayns::ApplicationParameters* a, ObjectHandler* h)
 {
     h->add_property("engine", &a->_engine, Flags::IgnoreRead | Flags::Optional);
+    h->add_property("plugins", &a->_plugins, Flags::Optional);
```
NadirRoGue, line 549: Here I think it would be appropriate to bitmask the flag Flags::IgnoreRead so that the clients cannot modify the list of plugins.

Adrien4193: Added 6ecc17e

NadirRoGue (approved):

PR #933, opened 2021-11-05
Frame export improvement and static global objects removal:

NadirRoGue:
- Frame export is now a core feature
- ODU camera definition is removed, and Frame export uses now the same camera and camera parameter definitions as set/get-camera and set/get-camera-params, allowing to use any kind of camera for frame exporting
- Fixed bug in NeuronColorHandler based color handlers
- Removed static global objects in circuit explorer


plugins/CircuitExplorer/plugin/io/morphology/neuron/NeuronBuilder.h
```diff
@@ -75,3 +75,42 @@ class NeuronBuilder
...
+ */
+class NeuronBuilderTable
+{
+public:
+    /**
+     * @brief NeuronBuilderTable constructor registers all available builders
+     */
+    NeuronBuilderTable();
+
+    /**
```
Adrien4193, line 91: Just a personal preference but usually I write it like that:
```
void registerBuilder(std::unique_ptr<NeuronBuilder> builder) { // register builder }

template<typename T, typename ...Args>
void register(Args &&...args)
{
    registerBuilder(std::make_unique<T>(std::forward<Args>(args)...));
}
```
Then you can register builders with no default constructors and it is always convenient to have a non-template method to do the registering (and the nasty SFINAE is not required as the code would not compile if T is not a Builder anyway).

Adrien4193 (approved):

PR #937, opened 2021-11-26
Moved ui, manipulators and input to BraynsViewer.

Adrien4193:


apps/BraynsViewer/ui/Application.cpp
```diff
@@ -77,6 +79,8 @@ void glfwErrorCallback(int error, const char* description)
 }
 } // namespace
 
+using namespace brayns;
```
NadirRoGue, line 82: Personal opinion, that might be problematic if new dependencies are added with similar names. It also make less clear from where certain types come from. But its not a big deal, so no need to change it.

NadirRoGue (dismissed):

apps/BraynsViewer/ui/Application.cpp
Adrien4193: Ah yeah actually I added that temporary but I forgot to remove it, thanks.

NadirRoGue (approved):

PR #938, opened 2021-11-26
Allow to build with GCC 11.

matz-e:
It seems that GCC has gotten a bit more pedantic about header inclusion,
and thus more parts of std:: have to be included directly.

Also deletes a new macro that interferes, and does not follow the
reserved namescheme.

Still has to be compiled with -Wno-error=range-loop-construct.

### DIFF
--- a/brayns/Brayns.cpp
+++ b/brayns/Brayns.cpp
@@ -57,7 +57,7 @@
 
 #include <brayns/pluginapi/PluginAPI.h>
 
-#include <mutex>  // std::unique_lock
+#include <mutex> // std::unique_lock
 #include <thread>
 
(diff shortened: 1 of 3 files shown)


matz-e:
Jenkins fails with some submodule initialization error. I don't think I can control that here, though…

NadirRoGue:
We disabled submodule processing because the develop branch, where we commit, does not have submodules anymore, and the CI was failing because apparently, attempting to clone submodules when there are none does not return with 0.
I've re-enabled them for this CI to pass, will disable them afterwards.

retest this please

NadirRoGue (approved):

PR #950, opened 2021-12-16
Fix compilation with GCC 11.

matz-e:
* Update filesystem to the latest from Github

  Failed with the usual range pedantic errors from GCC, but I could not
  manage to get this header included with -isystem due to the CMake
  setup.

* Sprinkle in missing headers that now are required after GCC cleaned
  up.


NadirRoGue (approved):
Approved.
Thanks again for the update. Soon we will get rid of filesystem.hpp since we are allowed to use c++17

PR #954, opened 2021-12-17
Brayns 313 - Convert export frames to task

NadirRoGue:
- New entrypoint 'export-frames', implements frame exporting as a task
- Moved FrameExporter class to network
- export-frames-to-disk and get-export-frames-progress entrypoint are kept, but deprecated (Will be removed on Brayns 3.0.0 release)


brayns/network/entrypoints/ExportFramesEntrypoint.cpp
```diff
@@ -0,0 +1,223 @@
...
+#include <brayns/utils/image/ImageEncoder.h>
+#include <brayns/utils/image/ImageFormat.h>
+
+#include <spdlog/fmt/fmt.h>
+
+namespace brayns
+{
+namespace
+{
+void logExportInfo(const ExportFramesParams& info)
```
Adrien4193, line 32: We can specialize fmt::formatter or the << operator of std::ostream with spdlog extension to do that (no need to change that it is fine also as we only use that once).

brayns/network/plugin/NetworkManagerEntrypoints.h
```diff
@@ -139,8 +140,11 @@ class NetworkManagerEntrypoints
         plugin.add<ClearLightsEntrypoint>();
         plugin.add<CancelEntrypoint>();
         plugin.add<RegistryEntrypoint>();
-        plugin.add<ExportFramesToDiskEntrypoint>();
-        plugin.add<GetExportFramesProgressEntrypoint>();
+        plugin.add<ExportFramesEntrypoint>();
+
+        auto frameExporter = std::make_shared<FrameExporter>();
```
Adrien4193, line 145: Could be unique ptr with a ref for the progress but it also make sense, no need to change that :).

Adrien4193 (approved):

brayns/network/entrypoints/ExportFramesEntrypoint.cpp
NadirRoGue: Ah good to know that. I think in this case is fine since its something used just here, but it would be useful in other cases, like specializing it for the Scene or the Model to easily log their status.

brayns/network/plugin/NetworkManagerEntrypoints.h
NadirRoGue: I thought it would be semantically more correct to use a shared_ptr so they can share ownership. 
It would be safe with the unique_ptr as well, since Its impossible get-export-frames-progress can be called after the export-frames-to-disk entrypoint is destroyed, which would result on a segfault. Nonetheless, shared_ptr will completely remove such far fetched case anyway.

Adrien4193: Yep you are right no need to make dangerous stuff for 1ns optimization

PR #956, opened 2021-12-17
Initial .gitlab-ci.yml script

NadirRoGue:


.gitlab-ci.yml
```diff
@@ -0,0 +1,105 @@
+stages:
+  - test
+  
+include:
+  - project: viz/ci/gitlabpipelines
+    file: index.yml
+  - project: hpc/gitlab-pipelines
+    file: spack-build-components.gitlab-ci.yml
+
```
matz-e, line 9: I would add this here (as default/tags is set specifically by `hpc/gitlab-pipelines`, this needs to be overriden)
```suggestion

default:
  tags: []
```

.gitlab-ci.yml
```diff
@@ -0,0 +1,105 @@
...
+  stage: test
+  
+  # TEST: TO BE REMOVED BEFORE MERGE!!!
+  rules:
+  #  - if: $CI_MERGE_REQUEST_IID
+    # run on default branch in order to upload coverage report
+    - if: $CI_COMMIT_BRANCH == "BRAYNS_306_Migrate_CI_to_Gitlab"
+  
+spack-build:
+  extends: .spack_setup_ccache
```
matz-e, line 88: And then reinstate it here again:
```suggestion
  tags: [bb5_map]
  extends: .spack_setup_ccache
```

.gitlab-ci.yml
```diff
@@ -0,0 +1,105 @@
...
+docker-build:
+  extends: .build-image-using-kaniko
+  stage: test
+  
+  # TEST: TO BE REMOVED BEFORE MERGE!!!
+  rules:
+  #  - if: $CI_MERGE_REQUEST_IID
+    # run on default branch in order to upload coverage report
+    - if: $CI_COMMIT_BRANCH == "BRAYNS_306_Migrate_CI_to_Gitlab"
+
```
matz-e, line 86: This is contained in the build-components… needs its tag set properly, too.
```suggestion
  
spack-setup:
  tags: [bb5_map]
```

matz-e (commented):
Sorry, GitLab YAML extension is a global-namespaced mess.

We set a default tag in the HPC pipelines that seems to get in the way of your docker builds. I've added some comments on how to override this and set it again for the spack jobs.

As an alternative, you could also spawn a separate pipeline for the HPC/Spack stuff, like here: https://bbpgitlab.epfl.ch/hpc/circuit-building/touchreader/-/blob/main/.gitlab-ci.yml

Since you don't want any `ctest` stuff back, that would be nicely isolated, and a workaround to missing namespaces.

NadirRoGue:
@matz-e Thanks for the help! ILl separate the spack build on a different pipline as you suggested

matz-e:
> @matz-e Thanks for the help! ILl separate the spack build on a different pipline as you suggested

Sorry this didn't work. I'm kind of puzzled, in the expanded YAML in the GitLab CI, it does not seem to reflect the latest changes (should also note that the Github vs GitLab sync sometimes lags up to ~5 min).

Separate pipelines make things clearer, though, even if they are more work.

.gitlab-ci.hpc-spack.yml
```diff
@@ -0,0 +1,28 @@
...
+  - project: hpc/gitlab-pipelines
+    file: spack-build-components.gitlab-ci.yml
+
+default:
+  tags: []
+
+spack-setup:
+  tags: [bb5_map]
+spack-build:
+  tags: [bb5_map]
```
matz-e, line 11:
```suggestion
spack-build:
```
This should be the default again, too

.gitlab-ci.hpc-spack.yml
```diff
@@ -0,0 +1,28 @@
...
+    SBATCH_ACCOUNT: proj3
+    SLURM_ACCOUNT: proj3
+    bb5_account: proj3
+    SPACK_PACKAGE_NAME: brayns
+    SPACK_PACKAGE_COMPILER: gcc
+    SPACK_PACKAGE_SPEC: ^ospray@1.8.5%intel +brion
+    DATADIR: /tmp 
+  
+  # TEST: TO BE REMOVED BEFORE MERGE!!!
+  rules:
```
matz-e, line 25: FWIW, these rules should then be located on the `worklfow` level or even in the parent pipeline, conditionally creating the subpipeline. Preferably the latter, as subpipelines will fail if there's no job created for them.

NadirRoGue: Thanks, changed!

NadirRoGue:
Hey @matz-e , thanks a lot for the help, now spack pipeline seems to be running, although it doesnt seem to be building brayns?

.gitlab-ci.hpc-spack.yml
```diff
@@ -0,0 +1,22 @@
...
+spack_setup:
+  extends: .spack_setup_ccache
+  variables:
+    SALLOC_ACCOUNT: proj3
+    SBATCH_ACCOUNT: proj3
+    SLURM_ACCOUNT: proj3
+    bb5_account: proj3
+    SPACK_PACKAGE_NAME: brayns
+    SPACK_PACKAGE_COMPILER: gcc
+    SPACK_PACKAGE_SPEC: ^ospray@1.8.5%intel +brion
```
matz-e, line 14: I think that these 3 variables should be defined on a file-global level, as both setup and build need them

NadirRoGue: Changed, thanks!

NadirRoGue:
@matz-e Definitely! It should be easy now to migrate any pipeline to gitlab!

.gitlab-ci.hpc-spack.yml
```diff
@@ -0,0 +1,49 @@
...
+   SPACK_PACKAGE_NAME: brayns
+   SPACK_PACKAGE_COMPILER: gcc
+   SPACK_PACKAGE_SPEC: +brion ^ospray@1.8.5%intel
+   DATADIR: /tmp 
+
+
+brayns-spack-build:
+  extends: .spack_build
+  stage: test
+  variables:
```
matz-e, line 21: These `variable` blocks could still be on the file level? Did that not work? Especially since they are repeated 3 times :)

NadirRoGue: Thanks!, Just made them file level to test it again (didnt really test them the previous time)
For the .ctest job, we are adding CTests in brayns, but seems like spack cannot find them. Is there any special way we have to set them up, or any variable I am not declaring for it?

NadirRoGue:
retest this please

PR #957, opened 2022-01-07
BRAYNS-323 Seg fault when trying to load a H5 file.

Adrien4193:


NadirRoGue (approved):
LGTM.

Apologies for the segfault caused by dangling references. I though I had fixed it already...

PR #967, opened 2022-01-24
BRAYNS-316 Remove assimp

Adrien4193:
Assimp replaced by custom mesh parsers (no libraries).

Supports files OBJ, PLY, STL and OFF.

No materials extracted, just a default white one for the whole mesh.

No triangulation (only supports triangular faces).

Vertex duplication for all formats (vertex count = index count).

An OBJ file can contain multiple objects.

A PLY file can have position, normal, color and texture coordinates (vertex + face) in vertices.

STL complete support.

Only positions supported in OFF files.


.gitlab-ci.viz.yml
```diff
@@ -1,16 +1,13 @@
 stages:
   - test
 
-
```
NadirRoGue, line 4: Why the removal of the blank lines? I added them for split clarity between jobs and env config.

brayns/io/loaders/MeshLoader.cpp
```diff
@@ -20,434 +20,189 @@
...
 
-namespace brayns
-{
 namespace
 {
-constexpr char LOADER_NAME[] = "mesh";
-constexpr float TOTAL_PROGRESS = 100.f;
-constexpr float LOADING_FRACTION = 50.f;
-constexpr float POST_LOADING_FRACTION = 50.f;
+using namespace brayns;
```
NadirRoGue, line 38: `using namespace` can lead to issues later, it is better to avoid them.

brayns/io/loaders/MeshLoader.h
```diff
@@ -20,53 +20,141 @@
...
+ * Vertices are duplicated (vertex count = 3 * index count).
+ *
+ * Only triangular faces are supported.
+ *
+ * Materials are not extracted.
+ *
+ * A default white material is used for the mesh model.
+ *
  */
 class MeshLoader : public Loader<MeshLoaderParameters>
```
NadirRoGue, line 106: Since the load configuration parameters are not used anymore, MeshLoader can extend `NoInputLoader` and avoid the `(void)parameters` on the source file?

NadirRoGue (approved):
LGTM. Left some minor comments

brayns/io/loaders/MeshLoader.h
NadirRoGue: Also we can get rid of the MeshLoaderParameters.h file

NadirRoGue (approved):

PR #973, opened 2022-01-26
Fix compilation:

NadirRoGue:
- Removes -w from Poco compilation. Replaced with a more precise warning turn off (-Wno-sign-compare)
- Added warnings turn off to CircuitExplorer to fix warnings from ospray (-Wno-strict-aliasing) and Brion (-Wno-deprecated-copy)
- Fixed all warnings/errors which appeared after disabling -w

### DIFF
--- a/brayns/io/Loader.h
+++ b/brayns/io/Loader.h
@@ -84,8 +84,10 @@ public:
      * @param const std::string& fileName Path to the file to queried
      * @param const std::string& extension Extracted file extension from fileName
      */
-    virtual bool isSupported(const std::string &, const std::string &extension) const
+    virtual bool isSupported(const std::string &fileName, const std::string &extension) const
     {
+        (void)fileName;
+
         const auto extensions = getSupportedExtensions();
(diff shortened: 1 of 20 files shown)


brayns/io/Loader.h
```diff
@@ -81,8 +81,10 @@ class AbstractLoader
      * @brief Query the loader if it can load the given file. Compares the file
      * extension to the supported extensions (removing the dot, if any, and
      * making them lowercase). Can be overriden to perform mor explicit checks
+     * @param const std::string& fileName Path to the file to queried
+     * @param const std::string& extension Extracted file extension from fileName
      */
-    virtual bool isSupported(const std::string &filename, const std::string &extension) const
+    virtual bool isSupported(const std::string &, const std::string &extension) const
```
Adrien4193, line 87: Simple personal preference but I prefer naming the unused parameters and adding a (void)parameter in the function body so we still see what it is.

NadirRoGue: Yes, that makes it more clear, Ill change that.

Adrien4193 (approved):

PR #1004, opened 2022-04-27
Updating copyright year 2022

bbpgithubaudit:


NadirRoGue:
Hello,

We do not commit to master, but to develop instead, where the copyright was updated on January. Thanks anyway for the PR.

PR #1005, opened 2022-04-29
BRAYNS 371 - Adding inspect (picking) functionality

NadirRoGue:


brayns/engine/Model.cpp
```diff
@@ -37,6 +37,12 @@ const std::map<std::string, std::string> &Model::getMetaData() const noexcept
     return _metadata;
 }
 
+void Model::onInspect(const Vector3f &hit, OSPGeometricModel modelHandle, uint32_t primitiveID, JsonObject &writeResult)
```
Adrien4193, line 40: Just some questions/remarks:

- I would put the 3 first arguments in a struct for flexibility.
- modelHandle is always the model the component is attached to right ? So we could access it without passing it in argument no ?
- Is there any reason why the result is an output parameter and not a return value ?

brayns/network/entrypoints/InspectEntrypoint.cpp
```diff
@@ -21,10 +21,103 @@
...
+class SceneInspector
+{
+public:
+    SceneInspector(brayns::Engine &engine, const brayns::Vector2f &inspectPosition)
+        : _engine(engine)
+        , _inspectPosition(inspectPosition)
+    {
+    }
+
+    brayns::InspectResult inspect()
```
Adrien4193, line 35: Just personal preference but I would put inspectPosition in inspect() parameters and call it just position.

(As engine is more a dependency and position the context of the function call).

brayns/network/entrypoints/InspectEntrypoint.cpp
```diff
@@ -21,10 +21,103 @@
...
+
+        const auto &hitPosition = result.position;
+        auto geometricModel = pickResult.model;
+        auto primitiveID = pickResult.primID;
+
+        brayns::JsonObject data;
+        model.onInspect(hitPosition, geometricModel, primitiveID, data);
+
+        auto &metadata = result.metadata;
+        metadata = brayns::JsonValue(data);
```
Adrien4193, line 63: The cast is really necessary ? Normally Poco::Var has a template constructor that accepts anything.

brayns/network/entrypoints/InspectEntrypoint.cpp
```diff
@@ -21,10 +21,103 @@
...
+        auto instanceIterator = std::find_if(
+            begin,
+            end,
+            [&](brayns::ModelInstance *instancePtr) { return instancePtr->handle() == hittedInstance; });
+
+        // Shouldn't happen, but..
+        if (instanceIterator == end)
+        {
+            throw brayns::JsonRpcException("Could not find the hitted model instance in the scene");
+        }
```
Adrien4193, line 106: I have defined a lot of standard exceptions in JsonRpcException.h, I think you can use InternalError or something like that.

plugins/CircuitExplorer/components/MorphologyCircuitComponent.cpp
```diff
@@ -101,6 +101,33 @@ void MorphologyCircuitComponent::onDestroy()
...
+
+    if (it == morphEnd)
+    {
+        return;
+    }
+
+    auto index = std::distance(morphBegin, it);
+
+    auto cellId = _ids[index];
+    writeResult.set("neuron id", brayns::JsonValue(cellId));
```
Adrien4193, line 128: Same here, cast is not necessary if cellId is an integer type.

Why do you call it "neuron id" and not "neuron_id" is it not supposed to be a JSON-RPC property ?

plugins/CircuitExplorer/components/SomaCircuitComponent.cpp
```diff
@@ -61,6 +61,23 @@ void SomaCircuitComponent::onDestroy()
...
+{
+    (void)hit;
+
+    if (modelHandle != _model)
+    {
+        return;
+    }
+
+    auto cellId = _ids[primitiveID];
+    writeResult.set("neuron id", brayns::JsonValue(cellId));
```
Adrien4193, line 78: Same remarks as for MorphologyComponent.

plugins/CircuitExplorer/components/SomaCircuitComponent.h
```diff
@@ -37,6 +37,12 @@ class SomaCircuitComponent final : public brayns::Component
 
     void onDestroy() override;
 
+    void onInspect(
+        const brayns::Vector3f &hit,
+        OSPGeometricModel modelHandle,
+        uint32_t primitiveID,
```
Adrien4193, line 43: As you forward it everywhere, I would definitively make a stuct...

Adrien4193 (approved):

brayns/network/entrypoints/InspectEntrypoint.cpp
```diff
@@ -21,10 +21,121 @@
...
+
+        auto &instance = _findHittedModel(pickResult.instance);
+        const auto inspectContext = _buildInspectContext(pickResult);
+        auto metadata = brayns::JsonObject();
+
+        auto &model = instance.getModel();
+        model.onInspect(inspectContext, metadata);
+
+        _releaseOSPRayHandles(pickResult);
```
Adrien4193, line 51: If findHittedModel throws (I know it shouldn't), this will never get called.

brayns/network/entrypoints/InspectEntrypoint.cpp
```diff
@@ -21,10 +21,121 @@
...
+        // Shouldn't happen, but..
+        if (instanceIterator == end)
+        {
+            throw brayns::InternalErrorException("Could not find the hitted model instance in the scene");
+        }
+
+        return **instanceIterator;
+    }
+
+    brayns::InspectContext _buildInspectContext(OSPPickResult &osprayPickResult)
```
Adrien4193, line 99: const OSPPickResult & ?

brayns/network/entrypoints/InspectEntrypoint.cpp
```diff
@@ -21,10 +21,121 @@
...
+        auto geometricModel = osprayPickResult.model;
+        auto primitiveIndex = osprayPickResult.primID;
+
+        return brayns::InspectContext{hitPosition, geometricModel, primitiveIndex};
+    }
+
+    brayns::InspectResult _buildResult(
+        const brayns::InspectContext &context,
+        const brayns::ModelInstance &instance,
+        brayns::JsonObject &metadata)
```
Adrien4193, line 112: const brayns::JsonObject & ?

Here making a move would be better (pass by copy and then move) but I think Poco::JSON::Object doesn't implement move construction / assignment (to be checked).

Adrien4193 (dismissed):
Some minor comments but you can merge it like that if you want.

Adrien4193 (approved):

PR #1010, opened 2022-05-12
Brayns 311 move parameters to common

Adrien4193:


plugins/CircuitExplorer/CircuitExplorerPlugin.cpp
```diff
@@ -22,8 +22,8 @@
 #include "CircuitExplorerPlugin.h"
 
 #include <brayns/common/Log.h>
+#include <brayns/common/parameters/ParametersManager.h>
```
NadirRoGue, line 25: I think I missed it in the last CircuitExplorer update, but seems like this include is not needed

NadirRoGue (approved):

NadirRoGue (approved):

PR #1019, opened 2022-05-18
Brayns 389 add task support to python

Adrien4193:


python/brayns/instance/client.py
```diff
@@ -33,9 +34,13 @@ def __init__(self, client: JsonRpcClient) -> None:
     def disconnect(self) -> None:
         self._client.disconnect()
 
-    def request(self, method: str, params: Any = None) -> Any:
```
NadirRoGue, line 36: I think it would be good to keep a method that blocks until the task is completed. It could just call task() and then wait for the future to complete to avoid repeating code, but that way, the clients who does not need async does not have to wait on the futures.

NadirRoGue: Nevermind, I saw that the request has been moved to instance now.

python/brayns/instance/jsonrpc/json_rpc_client.py
```diff
@@ -37,6 +38,9 @@ def __init__(self, websocket: WebSocket, logger: logging.Logger) -> None:
         self._handler = JsonRpcHandler(self._manager, logger)
         self._dispatcher = JsonRpcDispatcher(self._handler)
 
+    def __contains__(self, id: Union[int, str]) -> bool:
```
NadirRoGue, line 41: Just a personal preference, using the __contains__ to have the sintactic sugar of id in _client makes the client object to look like an id container, and does not make it obvious that you are checking for used message ids. I'd use a specific method that tell right away what is being check for.

However, is not an issue so you can merge.

NadirRoGue (approved):

python/brayns/instance/client.py
Adrien4193: It is already the case, but it is implemented in instance.py (base class of client). So I don't need to implement it for the mock.

python/brayns/instance/jsonrpc/json_rpc_client.py
Adrien4193: Yeah you are right I will do it in another PR.

PR #1040, opened 2022-06-29
Brayns 351 test api2

Adrien4193:


brayns/network/entrypoints/SnapshotEntrypoint.cpp
```diff
@@ -42,12 +42,11 @@ struct SnapshotResultHandler
     {
         const auto &formatName = imageSettings.getFormat();
         const auto format = brayns::string_utils::toLowercase(formatName);
```
NadirRoGue, line 44: Shouldn't we get the format now from the path?

python/brayns/core/common/bounds.py
```diff
@@ -36,6 +36,14 @@ def deserialize(message: dict) -> 'Bounds':
             max=Vector3(*message['max'])
         )
 
+    @classmethod
+    @property
+    def empty(cls) -> 'Bounds':
+        return cls(
+            min=Vector3.zero,
```
NadirRoGue, line 43: Just a remark: If the bounds class ever adds functionality to build bounds from points/other bounds, this would be a problem.

NadirRoGue (approved):

NadirRoGue (approved):

PR #1045, opened 2022-07-06
BRAYNS 396 - Restore opendeck camera

NadirRoGue:


README.md
```diff
@@ -16,6 +16,7 @@ Brayns also comes with some already-made plugins:
 * **CircuitInfo (Deprecated)**: A plugin which accepts queries to extract information from neuronal circuits.
 * **DTI**: A diffuse-tensor imaging visualization plugin.
 * **AtlasExplorer**: A plugin capable of interpreting and render NRRD volumes.
+* **CylindricCamera**: A plugin that adds a cylindric camera projection for inmersive environments.
```
Adrien4193, line 19: I would have called it OpenDeckPlugin maybe so if we need to develop other things for the OpenDeck, we can put it here.

brayns/engine/Camera.h
```diff
@@ -39,6 +40,12 @@ struct LookAt
     Vector3f up{0.f, 1.f, 0.f};
 };
 
+BRAYNS_JSON_ADAPTER_BEGIN(LookAt)
+BRAYNS_JSON_ADAPTER_ENTRY(position, "Position of the camera")
+BRAYNS_JSON_ADAPTER_ENTRY(target, "Target position at which the camera is looking")
+BRAYNS_JSON_ADAPTER_ENTRY(up, "Up vector to compute the camera orthonormal basis")
+BRAYNS_JSON_ADAPTER_END()
```
Adrien4193, line 47: Maybe we can put the brayns adapters / messages which are exposed to plugins in a separated folder at some point.
So the engine objects don't depend directly on JSON.

brayns/engine/Engine.cpp
```diff
@@ -218,6 +238,16 @@ Renderer &Engine::getRenderer() noexcept
     return *_renderer;
 }
 
+EngineObjectFactory<Camera> &Engine::getCameraFactory() noexcept
```
Adrien4193, line 241: Maybe using `EngineObjectFactory<Camera> = CameraFactory;` somewhere (and renderer).

brayns/engine/EngineObjectFactory.h
```diff
@@ -101,73 +100,28 @@ class EngineObjectFactory
...
+    template<typename SubT>
+    void registerType()
     {
-        if (_name.empty())
-        {
-            return _systemObject->clone();
-        }
-
-        return _factory->create(_name, _value);
+        auto name = SubT().getName();
```
Adrien4193, line 120: getName() could be static maybe ?

brayns/engine/renderers/ProductionRenderer.h
```diff
@@ -36,4 +37,25 @@ class ProductionRenderer final : public Renderer
...
+    getMaxRayBounces,
+    setMaxRayBounces,
+    "Max ray bounces per sample",
+    Required(false))
+BRAYNS_JSON_ADAPTER_GETSET(
+    "background_color",
+    getBackgroundColor,
+    setBackgroundColor,
+    "Background color",
+    Required(false))
```
Adrien4193, line 59: Maybe a macro with theses properties as they are defined for all renderers (but with only two it is fine like that also).

plugins/CylindricCamera/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+# This file is part of Brayns <https://github.com/BlueBrain/Brayns>
+
+list(APPEND CMAKE_MODULE_PATH ${OSPRAY_CMAKE_ROOT})
+include(ispc)
+
+file(GLOB_RECURSE BRAYNSCYLINDRICCAMERA_SOURCES RELATIVE ${CMAKE_CURRENT_LIST_DIR} *.cpp)
+file(GLOB_RECURSE BRAYNSCYLINDRICCAMERA_HEADERS RELATIVE ${CMAKE_CURRENT_LIST_DIR} *.h)
+file(GLOB_RECURSE BRAYNSCYLINDRICCAMERA_ISPC_SOURCES RELATIVE ${CMAKE_CURRENT_LIST_DIR} *.ispc)
+
+#list(APPEND BRAYNSCYLINDRICCAMERA_ISPC_SOURCES ${BRAYNSCYLINDRICCAMERA_SOURCES})
```
Adrien4193, line 14: Remove ?

plugins/CylindricCamera/CylindricCameraPlugin.cpp
```diff
@@ -0,0 +1,58 @@
...
+    ospray::Camera::registerType<OsprayCylindricCamera>(CylindricCamera::osprayName.c_str());
+
+    auto &engine = _api.getEngine();
+    auto &cameraFactory = engine.getCameraFactory();
+    cameraFactory.registerType<CylindricCamera>();
+}
+
+void CylindricCameraPlugin::registerEntrypoints(brayns::INetworkInterface &interface)
+{
+    auto &engine = _api.getEngine();
```
Adrien4193, line 47: As this plugin only uses engine, the brayns_create_plugin could extract it from the API and construct the plugin with only the engine (so the Plugin class only depends on engine).

plugins/CylindricCamera/camera/ospray/OsprayCylindricCameraShared.h
```diff
@@ -0,0 +1,52 @@
...
+        , dirDu(0.f)
+        , dirDv(0.f)
+        , imgPlaneSizeY(0.f)
+    {
+    }
+};
+} // namespace ispc
+#else
+};
+#endif // __cplusplus
```
Adrien4193, line 52: ISPC needs to compile these files as C and C++ sources ?

Adrien4193 (approved):

README.md
NadirRoGue: Even though its true we are going to use it for the opendeck, I made the camera configurable so that it can be used for any cylindrical projection (configurable fovy), I didnt want to mark it as something specific for the opendeck.

brayns/engine/Camera.h
NadirRoGue: Yes, that can be moved outside at a point.

brayns/engine/Engine.cpp
NadirRoGue: As personal preference I prefer not to hide the real type with `using`, unless is too long

brayns/engine/EngineObjectFactory.h
NadirRoGue: I though about an inline static variable, but I didnt implement it because the getName() is needed when calling get-camera-type, where you only have access to the abstract type.

brayns/engine/renderers/ProductionRenderer.h
NadirRoGue: Yes, I duplicated because they were only 2. The macro can be defined to avoid duplication.

plugins/CylindricCamera/camera/ospray/OsprayCylindricCameraShared.h
NadirRoGue: They are included in both C++ files and ispc files, so you need to disable C++ stuff for the ISPC compiler

Adrien4193 (approved):

Adrien4193 (approved):

PR #1047, opened 2022-07-12
Brayns 407 - Refactor BaseModel class out of system

NadirRoGue:


brayns/common/ModifiedFlag.h
```diff
@@ -0,0 +1,69 @@
...
+
+    ModifiedFlag(const ModifiedFlag &other)
+    {
+        (void)other;
+    }
+
+    ModifiedFlag &operator=(const ModifiedFlag &other)
+    {
+        (void)other;
+        return *this;
```
Adrien4193, line 42: So when an object is copied from another one we don't want to mark it as modified ?

brayns/common/Transform.cpp
```diff
@@ -56,11 +56,29 @@ const Vector3f &Transform::getScale() const noexcept
...
+    auto &rhsR = rhs.getRotation();
+    auto &rhsS = rhs.getScale();
+
+    return lhsT == rhsT && lhsR == rhsR && lhsS == rhsS;
+}
+
+bool operator!=(const Transform &lhs, const Transform &rhs)
+{
+    return !(lhs == rhs);
+}
```
Adrien4193, line 83: I would put that inside the Transform class and use rhs._translation directly.

For that we can also use std::tie(_translation, rotation, scal) == std::tie(other._translation, other.rotation, other.scale).

brayns/common/Transform.h
```diff
@@ -29,7 +28,7 @@ namespace brayns
  * @brief Defines the translation, rotation and scale parameters to be applied
  * to a scene asset.
  */
-class Transform : public BaseObject
+class Transform
```
Adrien4193, line 31: If it is not a BaseObject anymore, maybe we can make a struct with public members to lighten a bit the writing no ? (as you prefer)

brayns/common/parameters/SimulationParameters.cpp
```diff
@@ -81,4 +81,19 @@ void SimulationParameters::build(ArgvBuilder &builder)
...
+
+ModifiedFlag &SimulationParameters::getModifiedFlag() noexcept
+{
+    return _flag;
+}
+
+const ModifiedFlag &SimulationParameters::getModifiedFlag() const noexcept
+{
+    return _flag;
+}
```
Adrien4193, line 98: Like for the transfer function, can't we just have isModified() and resetModified() as public method ?

(So the flag is hidden as an implementation detail)

brayns/network/entrypoints/ExportFramesEntrypoint.cpp
```diff
@@ -141,7 +141,7 @@ class FrameExporter
         auto parametersManager = paramsManager;
         auto &simulation = parametersManager.getSimulationParameters();
         // Avoid the first frame not being triggered because the copy is in "not modified" state
-        simulation.markModified();
+        // simulation.markModified();
```
Adrien4193, line 144: Remove ?

python/brayns/core/light/quad_light.py
```diff
@@ -43,6 +43,6 @@ def emission_direction(self) -> Vector3:
     def serialize(self) -> dict:
         return self._to_dict({
             'position': list(self.bottom_left),
-            'edge1': list(self.edge2),
-            'edge2': list(self.edge1)
+            'edge1': list(self.edge1),
+            'edge2': list(self.edge2)
```
Adrien4193, line 47: Thanks :)

brayns/common/TransferFunction.cpp
```diff
@@ -83,4 +81,14 @@ Vector4f TransferFunction::getColorForValue(const float v) const
...
+
+ModifiedFlag &TransferFunction::getModifiedFlag() noexcept
+{
+    return _flag;
+}
+
+const ModifiedFlag &TransferFunction::getModifiedFlag() const noexcept
+{
+    return _flag;
+}
```
Adrien4193, line 93: The tf modified state cannot be handled at the upper level like the transform ?

So we can remove the modified flag from it and the transfer function only contains a minimal set of data.

And it could even be a struct with public field like the transform...

Adrien4193 (dismissed):

Adrien4193 (approved):

PR #1057, opened 2022-08-15
BRAYNS-423 Add functionalities to Python API.

Adrien4193:


python/brayns/utils/view/view.py
```diff
@@ -52,14 +52,23 @@ def deserialize(message: dict[str, Any]) -> View:
             up=Vector3(*message['up'])
         )
 
+    @property
+    def axis(self) -> Vector3:
```
NadirRoGue, line 56: Personally, I'd have called it view_vector (direction + magnitude).

NadirRoGue (approved):

PR #1061, opened 2022-08-26
Brayns 433 add movie to python cli

Adrien4193:


python/brayns/cli/core/camera_cli.py
```diff
@@ -44,23 +44,23 @@ def register(self, parser: argparse.ArgumentParser) -> None:
             '--camera_type',
             choices=[camera.name for camera in self.available],
             default=self.name,
-            help='Camera type',
+            help='Type of camera to use for render',
         )
         parser.add_argument(
```
NadirRoGue, line 49: Why not options to set the camera as position + target as well? Its the most common approach in computer graphics.

python/brayns/cli/core/service_cli.py
```diff
@@ -35,14 +35,14 @@ class ServiceCli(Cli):
 
     def register(self, parser: argparse.ArgumentParser) -> None:
         parser.add_argument(
-            '--port',
```
NadirRoGue, line 38: As a personal preference, I believe --port was clearer (and shorter), and the doc (if needed) already specifies that its the braynsService port.

python/brayns/cli/tasks/movie_cli.py
```diff
@@ -0,0 +1,91 @@
...
+        )
+        parser.add_argument(
+            '--fps',
+            type=float,
+            required=True,
+            metavar='RATE',
+            help='Movie FPS, should be the same as for export',
+        )
+        parser.add_argument(
+            '--ffmpeg_executable',
```
NadirRoGue, line 68: binary rather than executable? is shorter and more universal

python/tests/cli/core/test_movie_frames_cli.py
```diff
@@ -0,0 +1,84 @@
...
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+import unittest
+
+import brayns
+
+
+class TestLightCli(unittest.TestCase):
```
NadirRoGue, line 46: This doesnt seem like the correct name.

NadirRoGue: (or even the correct test, since there is another test frame exporter cli underneath)

NadirRoGue (approved):

PR #1062, opened 2022-08-26
BRAYNS 426 - Refactor inheritance out of engine objects

NadirRoGue:


brayns/engine/common/ArgumentInferer.h
```diff
@@ -0,0 +1,54 @@
...
+
+#pragma once
+
+namespace brayns
+{
+template<typename T>
+struct ArgumentInferer
+{
+    using argType = typename ArgumentInferer<decltype(&T::operator())>::argType;
+};
```
Adrien4193, line 29: The functorInfo trait I wrote in utils for the JSON stuff is already doing that no ?

brayns/engine/core/Engine.cpp
```diff
@@ -120,11 +104,10 @@ Engine::OsprayModuleHandler::~OsprayModuleHandler()
 Engine::Engine(ParametersManager &parameters)
     : _params(parameters)
     , _osprayDevice(OsprayDeviceInitializer::init(parameters))
-    , _camera(std::make_unique<PerspectiveCamera>())
-    , _renderer(std::make_unique<InteractiveRenderer>())
+    , _camera(Perspective())
+    , _renderer(Interactive())
```
Adrien4193, line 108: I would still call that PerspectiveProjection and InteractiveRenderer but if it is too long it's ok like that also.

NadirRoGue: I prefer this naming. Also for the renderer would be weird as the containing class is already called Renderer.

brayns/engine/common/ArgumentInferer.h
NadirRoGue: I will check, if so I can replace it now or in another pass to avoid having to retest everything.

brayns/engine/geometry/types/Capsule.cpp
```diff
@@ -0,0 +1,93 @@
...
+void GeometryTraits<Capsule>::updateData(ospray::cpp::Geometry &handle, std::vector<Capsule> &data)
+{
+    auto capsuleCount = data.size();
+
+    std::vector<uint32_t> indexData; //(capsuleCount);
+    indexData.reserve(capsuleCount);
+    for (uint32_t i = 0, index = 0; i < capsuleCount; ++i, index = index + 2)
+    {
+        indexData.push_back(index);
+        // indexData[i] = index;
```
Adrien4193, line 81: Remove ?

brayns/engine/geometry/types/Capsule.cpp
```diff
@@ -0,0 +1,93 @@
...
+    bounds.expand(matrix * Vector4f(p1Max, 1.f));
+    return bounds;
+}
+
+void GeometryTraits<Capsule>::updateData(ospray::cpp::Geometry &handle, std::vector<Capsule> &data)
+{
+    auto capsuleCount = data.size();
+
+    std::vector<uint32_t> indexData; //(capsuleCount);
+    indexData.reserve(capsuleCount);
```
Adrien4193, line 77: Remove comment ?

brayns/engine/geometry/Geometry.h
```diff
@@ -20,190 +20,110 @@
...
     template<typename Callable>
-    void forEach(const Callable &callback)
+    void forEach(Callable &&callback) noexcept
     {
-        _dirty = true;
-        const auto end = static_cast<uint32_t>(_primitives.size());
-        for (uint32_t i = 0; i < end; ++i)
+        using ArgType = DecayFirstArgType<Callable>; // typename ArgumentInferer<Callable>::argType;
+        auto cast = dynamic_cast<GeometryData<ArgType> *>(_data.get());
+        assert(cast);
```
Adrien4193, line 87: Remove comment ?

BTW if you want to throw if the cast fails you should use dynamic_cast<T &>.

And if the cast is not supposed to fail you can assert(dynamic_cast) and do a static_cast after so in release you don't have the cost of the RTTI (for micro-optimization :)).

brayns/engine/renderer/RendererTraits.h
```diff
@@ -0,0 +1,46 @@
...
+
+#include <string>
+
+namespace brayns
+{
+template<typename T>
+constexpr bool rendererSpecialized = false;
+
+template<typename T>
+class RendererTraits
```
Adrien4193, line 33: Trait should be singular no ? Or we can say it is all the traits of renderer and then it is fine...

Adrien4193 (approved):

brayns/engine/renderer/RendererTraits.h
NadirRoGue: Since each class contains more than 1 trait, its more correct to have the name in plural

brayns/engine/geometry/Geometry.h
NadirRoGue: Addressed. Really cool trick.

Adrien4193 (approved):

PR #1063, opened 2022-09-09
BRAYNS 438: Add new morphology geometry

NadirRoGue:
**This PR also fixes engine objects (camera/geometry/light/material/renderer/volume) move constructors and assigments:**
Because OSPRay move constructor/assigment are not declared noexcept, default noexcept move constructor/assigment on engine objects were not meeting the required constraints to be used by the standard library.
The options were to declared them without the noexcept specifier, or write them manually noexcept (It is safe since the OSPray object do not throw on move constructor/assigment).
The latter was choosen as this enables underneath stl operations that works with std::move_if_noexcept

### DIFF
--- a/brayns/engine/material/Material.cpp
+++ b/brayns/engine/material/Material.cpp
@@ -23,21 +23,6 @@
 
 namespace brayns
 {
-Material::Material(Material &&other) noexcept
-{
-    *this = other;
-}
-
-Material &Material::operator=(Material &&other) noexcept
-{
-    _handleName = std::move(other._handleName);
-    _materialName = std::move(other._materialName);
-    _handle = std::move(other._handle);
-    _data = std::move(other._data);
-    _flag = std::move(other._flag);
-    return *this;
-}
-
 Material::Material(const Material &other)
 {
     *this = other;
--- a/brayns/engine/material/Material.cpp
+++ b/brayns/engine/material/Material.cpp
@@ -23,6 +23,21 @@
 
 namespace brayns
 {
+Material::Material(Material &&other) noexcept
+{
+    *this = other;
+}
+
+Material &Material::operator=(Material &&other) noexcept
+{
+    _handleName = std::move(other._handleName);
+    _materialName = std::move(other._materialName);
+    _handle = std::move(other._handle);
+    _data = std::move(other._data);
+    _flag = std::move(other._flag);
+    return *this;
+}
+
 Material::Material(const Material &other)
 {
     *this = other;
--- a/brayns/engine/material/Material.cpp
+++ b/brayns/engine/material/Material.cpp
@@ -23,6 +23,21 @@
 
 namespace brayns
 {
+Material::Material(Material &&other) noexcept
+{
+    *this = other;
+}
+
+Material &Material::operator=(Material &&other) noexcept
+{
+    _handleName = std::move(other._handleName);
+    _materialName = std::move(other._materialName);
+    _handle = std::move(other._handle);
+    _data = std::move(other._data);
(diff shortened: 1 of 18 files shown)


brayns/engine/renderer/Renderer.cpp
```diff
@@ -34,6 +49,7 @@ Renderer &Renderer::operator=(const Renderer &other)
     _handle = ospray::cpp::Renderer(_handleName);
     _data = other._data->clone();
     _data->pushTo(_handle);
+    _flag = true;
```
Adrien4193, line 52: This looks a bit weird, I would do something like _flag.setModified(true).

brayns/engine/material/Material.cpp
```diff
@@ -35,6 +50,7 @@ Material &Material::operator=(const Material &other)
     _handle = ospray::cpp::Material("", _handleName);
     _data = other._data->clone();
     _data->pushTo(_handle);
+    _flag = true;
```
Adrien4193, line 53: This looks a bit weird, I would do something like _flag.setModified(true).

brayns/engine/volume/Volume.cpp
```diff
@@ -33,6 +48,7 @@ Volume &Volume::operator=(const Volume &other)
     _volumeName = other._volumeName;
     _handle = ospray::cpp::Volume(_handleName);
     _data = other._data->clone();
+    _flag = true;
```
Adrien4193, line 51: This looks a bit weird, I would do something like _flag.setModified(true).

plugins/CircuitExplorer/api/neuron/processors/SectionSmoother.cpp
```diff
@@ -0,0 +1,133 @@
...
+private:
+    float _somaRadiusMultiplier;
+    float _sectionRadiusChange;
+};
+} // namespace
+
+void SectionSmoother::process(NeuronMorphology &morphology) const
+{
+    constexpr float somaRadiusMultiplier = 0.25f;
+    constexpr float sectionRadiusChange = 0.11f;
```
Adrien4193, line 130: Actually I just found on YouTube that it is better to use static constexpr for constexpr locals so you don't push it on the stack on each function call.

Adrien4193 (approved):

plugins/CircuitExplorer/api/neuron/NeuronGeometryType.h
```diff
@@ -24,5 +24,6 @@ enum class NeuronGeometryType
 {
     Original,
     Smooth,
```
Adrien4193, line 26: Smooth is RadiusSmooth right ? Maybe we can rename it (if yes also in the Python API).

brayns/engine/renderer/Renderer.cpp
NadirRoGue: Will change on next PR

brayns/engine/material/Material.cpp
NadirRoGue: Will change on next PR

brayns/engine/volume/Volume.cpp
NadirRoGue: Will change on next PR

plugins/CircuitExplorer/api/neuron/processors/SectionSmoother.cpp
NadirRoGue: Good point, however these parameters will be soon made availble through the morphology loader, so Ill avoid the recommit.

PR #1070, opened 2022-09-29
BRAYNS 391 - Entity-component to Entity-component-systems

NadirRoGue:


brayns/common/ModifiedFlag.h
```diff
@@ -56,6 +56,11 @@ class ModifiedFlag
         return *this;
     }
 
+    void setModified(bool val) noexcept
```
Adrien4193, line 59: Abreviation, I would use bool modified.

brayns/engine/camera/Camera.cpp
```diff
@@ -100,6 +94,9 @@ bool Camera::commit()
     {
         return false;
     }
+    _data->pushTo(_handle);
+    _updateView();
+    _updateAspectRatio();
```
Adrien4193, line 99: So here you push the data to OSP before updating them ?

brayns/engine/common/InspectResultData.h
```diff
@@ -0,0 +1,28 @@
...
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#pragma once
+
+#include <brayns/json/JsonType.h>
+
+namespace brayns
+{
+using InspectResultData = JsonObject;
```
Adrien4193, line 27: What the point of that ?

brayns/engine/common/SimulationScanner.cpp
```diff
@@ -45,43 +45,41 @@ struct SimulationUpdater
...
-        auto startTime = simulation->getStartTime();
-        auto endTime = simulation->getEndTime();
-        auto dt = simulation->getDT();
+        auto startTime = simulation->startTime;
+        auto endTime = simulation->endTime;
+        auto dt = simulation->dt;
 
         earlierStart = startTime < earlierStart ? startTime : earlierStart;
         latestEnd = endTime > latestEnd ? endTime : latestEnd;
         smallestDt = dt < smallestDt ? dt : smallestDt;
```
Adrien4193, line 74: std::min and std::max maybe ?

brayns/engine/components/LoadInformation.h
```diff
@@ -18,24 +18,26 @@
...
+#include <string>
 
 namespace brayns
 {
-MaterialComponent::MaterialComponent()
-    : _material(Phong())
+struct LoadInformation
 {
-}
+    enum class LoadType
```
Adrien4193, line 31: Just personal preference but I would call that LoadSource {Unknown, File, Blob} and put it outside the class so it is Unknown if not set (default value) and we don't need to do LoadInformation::LoadType to access it.

brayns/engine/model/Components.h
```diff
@@ -0,0 +1,158 @@
...
+    {
+        static_assert(std::is_same_v<std::decay_t<T>, T>, "Only decayed components allowed");
+
+        auto &entry = _components.emplace_back(typeid(T), std::make_unique<Component<T>>());
+        auto &component = static_cast<Component<T> &>(*entry.component);
+        auto &data = component.data;
+
+        if constexpr (std::is_aggregate_v<T>)
+        {
+            data = T{std::forward<Args>(args)...};
```
Adrien4193, line 45: This one works also for non-aggregates no ?

brayns/engine/model/Components.h
```diff
@@ -0,0 +1,158 @@
...
+    {
+        auto it = _findIterator<T>();
+        if (it == _components.end())
+        {
+            return nullptr;
+        }
+        auto &entry = *it;
+        auto &component = entry.component;
+        auto &cast = static_cast<Component<T> &>(*component);
+        return &cast.data;
```
Adrien4193, line 106: As the component is a pointer, it can be retreived as a non-const in both find versions (const and non-const) so we could do this to avoid duplications:

T *find() {return _find() }

const T *find() const {return _find() }

T *_find() const { /*block of code above*/}

brayns/engine/model/Components.h
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    template<typename T>
+    bool has() const
+    {
+        return find<T>() != nullptr;
+    }
```
Adrien4193, line 113: if model.find<T>() is the same as if model.has<T>() no ?

brayns/engine/model/ModelInstance.cpp
```diff
@@ -75,19 +74,8 @@ const Bounds &ModelInstance::getBounds() const noexcept
...
-
-    // In the event that the model used by this instance has no geometry,
-    // The result of computeBounds() would be bounds with min > max.
-    const auto &min = _bounds.getMin();
-    const auto &max = _bounds.getMax();
-    if (glm::max(min, max) == min)
-    {
-        _bounds = Bounds(Vector3f(0.f), Vector3f(0.f));
-    }
+    Log::debug("[ModelInstance {}] Computing bounds", _instanceID);
```
Adrien4193, line 77: I don't think it is the responsibility of the model instance to log stuff but more for the caller at a higher level but not a big deal...

brayns/engine/model/ModelInstance.h
```diff
@@ -102,27 +106,28 @@ class ModelInstance
...
     /**
      * @brief Commit implementation
      */
     bool commit();
 
 private:
-    const uint32_t _modelInstanceID{};
-    Model &_model;
+    /**
+     * @brief Returns the transform matrix obtained by multiplying the underlying Model base trasnform (if any)
```
Adrien4193, line 118: transform typo

brayns/engine/model/Systems.cpp
```diff
@@ -0,0 +1,75 @@
...
+    return InspectResultData();
+}
+
+Bounds Systems::computeBounds(const Matrix4f &matrix, Components &components)
+{
+    if (_bounds)
+    {
+        return _bounds->compute(matrix, components);
+    }
+    return {};
```
Adrien4193, line 73: Default constructor is infinite bounds, not empty, is it intended?

brayns/engine/scene/ModelManager.cpp
```diff
@@ -79,24 +79,22 @@ class ModelFinder
...
 {
-ModelInstance &ModelManager::addModel(ModelLoadParameters params, std::unique_ptr<Model> model)
+ModelInstance *ModelManager::addModel(std::unique_ptr<Model> model)
 {
     std::vector<std::unique_ptr<Model>> models;
     models.push_back(std::move(model));
-    auto instances = addModels(std::move(params), std::move(models));
-    return *(instances.front());
+    auto instances = addModels(std::move(models));
+    return instances.front();
```
Adrien4193, line 87: I would do the inverse, having addModels calling addModel and avoid creating the vector for single model.

brayns/engine/scene/ModelsOperations.cpp
```diff
@@ -18,33 +18,41 @@
...
+    for (auto instance : instances)
     {
-        return false;
+        auto &model = instance->getModel();
+        auto &components = model.getComponents();
+        if (components.has<Lights>() || components.has<ClipperViews>())
+        {
+            continue;
+        }
+        ids.push_back(instance->getID());
```
Adrien4193, line 53: Can't we call directly removeInstance(id) instead of building a vector ? Is it because the bounds are recomputed automatically ?

brayns/engine/scene/ModelsOperations.cpp
```diff
@@ -18,33 +18,41 @@
...
-ColorRamp &ExtractComponent::colorRamp(Model &model)
+void ModelsOperations::removeClippers(ModelManager &models)
 {
-    auto &component = model.getComponent<ColorRampComponent>();
-    return component.getColorRamp();
+    removeModelsWithComponent<ClipperViews>(models);
 }
 
-bool ExtractComponent::simulationEnabled(Model &model)
+void ModelsOperations::removeObjects(ModelManager &models)
```
Adrien4193, line 38: Is it removeRenderable equivalent ?

brayns/engine/scene/ModelsOperations.cpp
```diff
@@ -18,33 +18,41 @@
...
+
+    std::vector<uint32_t> ids;
+    ids.reserve(instances.size());
+
+    for (auto instance : instances)
     {
-        return false;
+        auto &model = instance->getModel();
+        auto &components = model.getComponents();
+        if (components.has<Lights>() || components.has<ClipperViews>())
```
Adrien4193, line 49: It will not scale very well if we need to add new non-renderable types...

Could it be something like removeModelsWithComponent<Renderable>(models); ?

brayns/engine/scene/ModelsOperations.h
```diff
@@ -20,42 +20,38 @@
...
+        std::vector<uint32_t> ids;
+        ids.reserve(instances.size());
+
+        for (auto instance : instances)
+        {
+            auto &model = instance->getModel();
+            auto &components = model.getComponents();
+            if (components.has<Component>())
+            {
+                ids.push_back(instance->getID());
```
Adrien4193, line 46: Same question as in the cpp but it is not a big deal if it is simpler like that.

brayns/engine/scene/ModelsOperations.h
```diff
@@ -20,42 +20,38 @@
...
 {
 public:
-    template<typename T>
-    ClippingComponent(T primitive)
-        : ClippingComponent(std::vector<T>{std::move(primitive)})
+    template<typename Component>
+    static void removeModelsWithComponent(ModelManager &models)
     {
-    }
+        static_assert(std::is_same_v<Component, std::decay_t<Component>>, "Only decayed types allowed");
```
Adrien4193, line 33: This check should be at the model level (~component manager) no ?

Otherwise it should be made everywhere we use components no ?

If you need to make this check at other places maybe a type trait like IsValidComponentType could be useful ?

brayns/engine/systems/GenericBoundsSystem.h
```diff
@@ -20,31 +20,36 @@
...
+            Bounds local;
 
-    virtual Bounds computeBounds(const Matrix4f &transform) const noexcept override;
-    virtual void onCreate() override;
-    bool commit() override;
+#pragma omp for
+            for (size_t i = 0; i < elements.size(); ++i)
+            {
+                auto &element = elements[i];
+                local.expand(element.computeBounds(matrix));
```
Adrien4193, line 45: I am not very familiar with OMP but here we have potentially multiple threads touching local no ?

thread1: for i in range(0, N) local.expand()
thread2: for i in range(N, M) local.expand()
...

brayns/io/loaders/protein/ProteinLoader.cpp
```diff
@@ -19,440 +20,223 @@
...
-                {
-                    atom.radius = atomic_radii[i].radius;
-                    found = true;
-                }
-                ++i;
-            }
+    static Atom _processLine(std::string_view line)
+    {
+        auto tokens = brayns::StringSplitter::splitTokens(line);
+        assert(tokens.size() >= 12);
```
Adrien4193, line 114: It could be done without allocation using StringCount::countTokens and StringExtractor::extractToken but it is fine like that also.

But I wouldn't use assert here and throw an exception as we want to check the input files even in release mode...

Otherwise the index access will crash...

brayns/engine/common/InspectResultData.h
NadirRoGue: Giving a meaninful name to what the JsonObject will be used

brayns/engine/model/Components.h
NadirRoGue: Yes, but the purpose of the condition is to be able to pass directly struct fields in add()

brayns/network/entrypoints/ModelMaterialEntrypoint.h
```diff
@@ -134,83 +141,89 @@ class GetMaterialEntrypoint : public Entrypoint<GetModelMessage, MaterialType>
...
-        if (auto cast = material.as<MaterialType>())
+        auto &components = model.getComponents();
+
+        auto material = components.find<Material>();
+        if (!material)
+        {
+            throw JsonRpcException("The model does not have material");
+        }
+
+        if (auto cast = material->as<MaterialType>())
```
Adrien4193, line 163: Just as "common practice", I would do all the checks before doing the actual work (even if it takes one more line) but you can leave it like that.

brayns/engine/model/Components.h
NadirRoGue: It is just syntactic sugar to be more explicit

brayns/engine/model/ModelInstance.cpp
NadirRoGue: I agree. Furthermore this was used just to debug the first refactor and it is not needed anymore.

brayns/engine/model/Systems.cpp
NadirRoGue: Yes. Otherwise, a bounds with min (0, 0, 0) and max (0, 0, 0) will have affect a composition of all instance bounds

brayns/engine/scene/ModelsOperations.cpp
NadirRoGue: That would result in modifying the same vector we are iterating over

brayns/engine/scene/ModelsOperations.cpp
NadirRoGue: Yes, I can rename it already to renderable if you think thats more appropiate

brayns/engine/scene/ModelsOperations.h
NadirRoGue: It is made during the component add (so only non decayed types can be searched, if you search for a non decayed type it will simply tell you it doesnt have it). In this part it is probably not necessary, following the same logic (trying to remove something that is not there)

brayns/engine/systems/GenericBoundsSystem.h
NadirRoGue: No, the parallelization block starts at `#pragma omp parallel` . `Bounds local;` is a local variable to each thread. Then we merge all together with a mutex section

brayns/common/ModifiedFlag.h
NadirRoGue: Addressed

brayns/engine/common/SimulationScanner.cpp
NadirRoGue: Addressed

brayns/engine/common/InspectResultData.h
NadirRoGue: I will actually add an struct around it to create a real type so that it cannot collide with other components.

brayns/engine/components/LoadInformation.h
NadirRoGue: Addressed

brayns/engine/model/Components.h
NadirRoGue: Addressed

brayns/engine/model/ModelInstance.h
NadirRoGue: Addressed

brayns/engine/scene/ModelManager.cpp
NadirRoGue: Addressed

brayns/engine/scene/ModelsOperations.cpp
NadirRoGue: Addressed

brayns/io/loaders/protein/ProteinLoader.cpp
NadirRoGue: Addressed

brayns/network/entrypoints/ModelMaterialEntrypoint.h
NadirRoGue: Addressed

plugins/CircuitExplorer/io/sonataloader/populations/nodes/common/NeuronReportFactory.cpp
```diff
@@ -82,53 +82,123 @@ struct SonataCompartmentMapping
...
+public:
+    static bool hasReport(const sl::NodeLoadContext &context)
+    {
+        auto &params = context.params;
+        auto reportType = params.report_type;
+        if (reportType == sl::ReportType::None)
+        {
+            return false;
+        }
+        return true;
```
Adrien4193, line 169: return reportType != ReportType::None;

NadirRoGue: Good catch, addressed.

plugins/DTI/systems/SpikeReportSystem.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+/**
+ * @brief Check the status of the simulation (enabled, disabled, restore defaults, etc.)
+ */
+class SimulationChecks
+{
+public:
+    static bool enabled(const brayns::SimulationInfo &info, dti::SpikeReportData &data, brayns::Components &components)
+    {
+        if (!info.enabled)
```
Adrien4193, line 108: No need to change that but we could use guard clause instead of nested conditions.

Something like this:

if info.enabled: return true
if data.lastEnabledFlag: paint + reset
return false

python/brayns/core/clipping/add_clipping_geometry.py
```diff
@@ -20,19 +20,21 @@
...
 from brayns.network import Instance
 
+from ..model import Model, deserialize_model
 from .clipping_geometry import ClippingGeometry
 
 
-def add_clipping_geometry(instance: Instance, geometry: ClippingGeometry) -> int:
-    """Add a clipping geometry to the given instance and return its ID.
+def add_clipping_geometry(instance: Instance, geometry: ClippingGeometry) -> Model:
+    """Add a clipping geometry to the given instance and return the model.
```
Adrien4193, line 28: "its model" maybe ?

python/brayns/core/light/add_light.py
```diff
@@ -21,18 +21,20 @@
 from brayns.network import Instance
 
 from .light import Light
+from ..model import Model, deserialize_model
 
 
-def add_light(instance: Instance, light: Light) -> int:
-    """Add a given light to an instance and return its ID.
+def add_light(instance: Instance, light: Light) -> Model:
+    """Add a given light to an instance and return the model.
```
Adrien4193, line 28: "its model" ?

plugins/DTI/systems/SpikeReportSystem.cpp
NadirRoGue: Actually I did it like that on the radiireportsystem, and I prefer it as well that way. Addressed.

python/brayns/service/plugin.py
```diff
@@ -33,7 +33,7 @@ class Plugin(Enum):
     CIRCUIT_EXPLORER = 'braynsCircuitExplorer'
     ATLAS_EXPLORER = 'braynsAtlasExplorer'
     CYLINDRIC_CAMERA = 'braynsCylindricCamera'
-    CIRCUIT_INFO = 'braynsCircuitInfo'
+    #CIRCUIT_INFO = 'braynsCircuitInfo'
```
Adrien4193, line 36: Better leaving it uncommented so we don't forget to remove it when we will remove the plugin no ?

python/testapi/core/geometry/test_add_planes.py
```diff
@@ -30,7 +30,6 @@ def test_add_planes(self) -> None:
             brayns.Plane(4, 5, 6, 7)
         ])
         self.assertEqual(model.id, 0)
-        self.assertEqual(model.bounds, brayns.Bounds.empty)
-        self.assertEqual(model.metadata, {})
+        self.assertEqual(model.info, {})
         self.assertEqual(model.visible, True)
```
Adrien4193, line 34: Why removing the test of the bounds ?

python/brayns/core/clipping/add_clipping_geometry.py
NadirRoGue: Addressed

python/brayns/core/light/add_light.py
NadirRoGue: Addressed

python/tests/core/clipping/test_add_clipping_geometries.py
```diff
@@ -22,15 +22,16 @@
...
 class TestAddClippingGeometries(unittest.TestCase):
 
     def test_add_clipping_geometry(self) -> None:
-        instance = MockInstance(0)
+        instance = MockInstance(MockModel.message)
         plane = brayns.ClipPlane(1, 2, 3, 4)
-        test = brayns.add_clipping_geometry(instance, plane)
-        self.assertEqual(test, 0)
+        test_model = brayns.add_clipping_geometry(instance, plane)
+        self.assertEqual(test_model, MockModel.model)
```
Adrien4193, line 34: In the add_light tests you left it named "model", better to stay coherent.

python/brayns/service/plugin.py
NadirRoGue: Yes, I forgot to renable it. Addressed.

tests/helpers/BraynsTestUtils.h
```diff
@@ -42,17 +45,29 @@ class BraynsTestUtils
     {
         auto &engine = brayns.getEngine();
         auto &scene = engine.getScene();
-        scene.addLight(std::move(light));
+        auto &models = scene.getModels();
+
+        auto model = std::make_unique<brayns::Model>();
+
+        auto &components = model->getComponents();
+        auto &lights = components.add<brayns::Lights>(); //(std::vector<brayns::Light>{std::move(light)});
```
Adrien4193, line 53: Comment to remove.

python/testapi/core/geometry/test_add_planes.py
NadirRoGue: It is comparing it to bounds with min (0, 0, 0) and max (0, 0, 0). The rendered returned the same, but thats an error, since the must return infinite bounds to avoid interference with bound testings (i.e., the renderer now returns min(+inf), max(-inf) for invalid bounds, so the test was failing)

python/tests/core/clipping/test_add_clipping_geometries.py
NadirRoGue: Addressed

tests/helpers/BraynsTestUtils.h
NadirRoGue: Addressed

Adrien4193 (approved):

PR #1072, opened 2022-09-30
BRAYNS 450 - Make engine object fields non mandatory

NadirRoGue:


brayns/network/adapters/ColorRampAdapter.h
```diff
@@ -28,7 +28,7 @@
 namespace brayns
 {
 BRAYNS_JSON_ADAPTER_BEGIN(ColorRamp)
-BRAYNS_JSON_ADAPTER_GETSET("range", getValuesRange, setValuesRange, "Values range")
-BRAYNS_JSON_ADAPTER_GETSET("colors", getColors, setColors, "List of colors (RGBA) to map")
+BRAYNS_JSON_ADAPTER_GETSET("range", getValuesRange, setValuesRange, "Values range", Required(false))
+BRAYNS_JSON_ADAPTER_GETSET("colors", getColors, setColors, "List of colors (RGBA) to map", Required(false))
```
Adrien4193, line 32: Get / Set are not mandatory by default (so when the field is not present, the function is not called) but it is ok like that also.

Adrien4193 (approved):

brayns/network/adapters/ColorRampAdapter.h
NadirRoGue: Yes, I noticed after a while. Since its not harmful, to avoid retriggering the CI, it can be removed on another PR

PR #1074, opened 2022-09-30
Brayns 435 - Use core string api on atlas explorer

NadirRoGue:


plugins/AtlasExplorer/io/nrrdloader/data/decoders/common/Tokenizer.cpp
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+#include "Tokenizer.h"
+
+#include <brayns/utils/string/StringCounter.h>
+#include <brayns/utils/string/StringExtractor.h>
+
+namespace
+{
+inline static const std::string_view delimiters = " /r/v/f/t/n";
```
Adrien4193, line 28: I would put the delimiters as a static inline string_view in StringInfo (brayns/utils/string) so we can reuse it (I have seen some duplicates of this in the code).

plugins/AtlasExplorer/io/nrrdloader/header/HeaderParser.cpp
```diff
@@ -29,71 +33,59 @@
 
 namespace
 {
+class LineBreakFinder
+{
+public:
+    static std::string fromView(std::string_view view)
```
Adrien4193, line 39: Could return a string view.

Adrien4193 (approved):

PR #1082, opened 2022-10-17
BRAYNS 456 - Added bounded plane geometry

NadirRoGue:


python/brayns/core/geometry/bounded_plane.py
```diff
@@ -0,0 +1,79 @@
...
+# details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from multiprocessing.sharedctypes import Value
```
Adrien4193, line 24: VSCode autocompletion overdid that it seems :)

python/brayns/core/geometry/bounded_plane.py
```diff
@@ -0,0 +1,79 @@
...
+    :type bounds: Box
+    """
+
+    a: float
+    b: float
+    c: float
+    d: float
+    bounds: Box
+
+    def __new__(cls, a: float, b: float, c: float, d: float, bounds: Box) -> BoundedPlane:
```
Adrien4193, line 58: You can override __post_init__ like this:

```
def __post_init__(self) -> None:
    if ... raise ValueError()
```

Overriding new is only for very specific cases (mostly when inheriting from built-in types.

The advantage of __post_init__ is that you don't need to duplicate the a, b, c, d, bounds and to call the __new__.

python/brayns/core/geometry/bounded_plane.py
```diff
@@ -0,0 +1,79 @@
...
+        diff = bounds.max - bounds.min
```
Adrien4193, line 59: bounds.size if using Bounds.

python/brayns/core/geometry/bounded_plane.py
NadirRoGue: __new__ allows me to throw the exception before the object is actually built, thats why I used it.

python/brayns/core/geometry/bounded_plane.py
```diff
@@ -0,0 +1,79 @@
...
+    :type d: float
+    :param bounds: Axis aligned bounds to limit the plane.
+    :type bounds: Box
+    """
+
+    a: float
+    b: float
+    c: float
+    d: float
+    bounds: Box
```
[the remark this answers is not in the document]
NadirRoGue: I need serialization, which Bounds lacks

python/brayns/core/geometry/bounded_plane.py
NadirRoGue: I wont use Bounds, because I need serialization.

brayns/engine/geometry/types/BoundedPlane.cpp
```diff
@@ -0,0 +1,56 @@
...
+namespace brayns
+{
+Bounds GeometryTraits<BoundedPlane>::computeBounds(const Matrix4f &matrix, const BoundedPlane &data)
+{
+    return GeometryTraits<Box>::computeBounds(matrix, data.bounds);
+}
+
+void GeometryTraits<BoundedPlane>::updateData(ospray::cpp::Geometry &handle, std::vector<BoundedPlane> &data)
+{
+    constexpr auto stride = sizeof(float) * 10;
```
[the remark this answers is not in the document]
NadirRoGue: It does work. I will change it on the next PR.

python/tests/core/geometry/test_bounded_plane.py
```diff
@@ -0,0 +1,53 @@
...
+        bounded_plane = brayns.BoundedPlane(
+            a=0,
+            b=0,
+            c=1,
+            d=0,
+            bounds=brayns.Box(
+                min=brayns.Vector3.zero,
+                max=brayns.Vector3.one,
+            )
+        )
```
Adrien4193, line 41: Just a suggestion, but to spare some size :)

```
plane = brayns.BoundedPlane(0, 0, 1, 0, bounds=Box(
   min=...
   max=...
))
```

Adrien4193 (approved):

PR #1083, opened 2022-10-17
Brayns 455 add tests for async entrypoints

Adrien4193:


python/brayns/network/future.py
```diff
@@ -0,0 +1,65 @@
...
+    def has_progress(self) -> bool:
+        return self._future.has_progress()
+
+    def get_progress(self) -> JsonRpcProgress:
+        return self._future.get_progress()
+
+    def wait_for_reply(self) -> JsonRpcReply:
+        return self._future.wait_for_reply()
+
+    def wait_for_result(self) -> T:
```
NadirRoGue, line 57: Personally, I would call these functions `get_reply()` and `get()` or `get_result()`, as typically on async APIs `wait()` functions does not return, but just wait for the result to be available.

python/brayns/network/instance.py
```diff
@@ -22,7 +22,7 @@
 
 from typing import Any, Protocol, TypeVar
 
-from .jsonrpc import JsonRpcFuture, JsonRpcRequest
+from .jsonrpc import JsonRpcFuture, JsonRpcReply, JsonRpcRequest
```
NadirRoGue, line 25: Just as curiosity (and its not from this PR), Reply is used for human to human communication. Response is more correct for communication with machines.

python/testapi/core/render/test_exporter.py
```diff
@@ -68,5 +86,10 @@ def _check_frames(self) -> None:
                 validator.validate_file(test, ref)
             except RuntimeError as e:
                 errors.append(str(e))
-        self.output.rmdir()
+        self._cleanup()
         self.assertFalse(errors)
+
+    def _cleanup(self) -> None:
```
NadirRoGue, line 92: I think it would be safer to use ${TMPDIR} if setted (or /tmp if the former is not set) than manually deleting files and folders.

NadirRoGue (approved):

PR #1089, opened 2022-10-20
BRAYNS 440 - Simplify model management

NadirRoGue:


brayns/engine/scene/ModelManager.cpp
```diff
@@ -70,49 +41,53 @@ class ModelFinder
     }
 
     template<typename Container>
-    static auto &findInstance(Container &instances, const uint32_t instanceId)
+    static auto &find(Container &instances, uint32_t instanceId)
```
Adrien4193, line 44: I would call that get as it returns a reference and use find when the result can be null.

Adrien4193 (approved):

PR #1132, opened 2022-12-14
Fix header includes for GCC 12.

matz-e:
These headers were probably implicitly included before, but no longer in GCC 12. Comment with the component used where not obvious.


matz-e:
Superseded by #1133

PR #1139, opened 2023-01-09
BRAYNS 496 - Restore volume loaders

NadirRoGue:


brayns/io/loaders/volume/VolumeLoader.cpp
```diff
@@ -19,182 +19,222 @@
...
+public:
+    static std::unordered_map<std::string, std::string> parse(const std::string &filename)
+    {
+        auto content = brayns::FileReader::read(filename);
+        auto view = std::string_view(content);
+
+        auto result = std::unordered_map<std::string, std::string>();
+
+        auto ctr = 1ul;
+        while (!view.empty())
```
Adrien4193, line 50: I had written a FileStream class to combine iteration with line number just in case.

brayns/io/loaders/volume/VolumeLoader.cpp
```diff
@@ -19,182 +19,222 @@
...
+            result.emplace(key, value);
+        }
 
-template<typename T>
-glm::vec<3, T> parseVec3(const std::string &str, std::function<T(std::string)> conv)
+        return result;
+    }
+};
+
+class MhdRequiredKeys
```
Adrien4193, line 70: Just personal preference but I would parse the header as a struct if all these keys are mandatory and raise the errors from here.

brayns/io/loaders/volume/VolumeLoader.cpp
```diff
@@ -207,50 +247,40 @@ std::vector<std::string> RawVolumeLoader::getSupportedExtensions() const
     return {"raw"};
```
Adrien4193, line 247: Useless braces.

brayns/io/loaders/volume/VolumeLoader.cpp
```diff
@@ -207,50 +247,40 @@ std::vector<std::string> RawVolumeLoader::getSupportedExtensions() const
...
     params.dimensions = dimensions;
     params.spacing = spacing;
-    params.type = type;
+    params.data_type = dataType;
 
-    return RawVolumeLoader().importFromFile(volumeFile, callback, params, scene);
+    return RawVolumeLoader().importFromFile(volumeFilePath, callback, params);
 }
 
 std::string MHDVolumeLoader::getName() const
```
Adrien4193, line 286: I know it is the old name but we have different case for almost all loader names ^^.
Like SONATA loader, mesh, mhd-volume,...

python/brayns/core/loader/mhd_volume_loader.py
```diff
@@ -0,0 +1,43 @@
...
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+from typing import Any
+
+from .loader import Loader
+
+
+class MhdVolumeLoader(Loader):
```
Adrien4193, line 25: For the other loaders I put the supported extensions as a ClassVar[str] so we can see from the Python API what is supported and we don't make typo in hard-coded strings.

python/brayns/core/loader/raw_volume_loader.py
```diff
@@ -0,0 +1,62 @@
...
+    :type dimensions: Vector3.
+    :param spacing: World-space dimensions of a voxel.
+    :type size: Vector3.
+    :param data_type: Type of data to interpret the volume bytes.
+    :type data_type: str.
+    """
+
+    dimensions: Vector3
+    spacing: Vector3
+    data_type: str
```
Adrien4193, line 44: I would make an enum for that.

python/brayns/core/loader/volume_data_type.py
```diff
@@ -0,0 +1,31 @@
...
+# details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+from typing import ClassVar
+
+
+class VolumeDataType:
```
Adrien4193, line 23: Maybe enum would be better so we see from the loader where the list of available values are and it is strongly typed.

python/tests/core/loader/test_mhd_volume_loader.py
```diff
@@ -0,0 +1,27 @@
...
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+import unittest
+
+import brayns
+
+
+class TestMhdVolumeLoader(unittest.TestCase):
```
Adrien4193, line 25: I would put the test_get_properties and assert empty object just to be homogeneous.

Adrien4193 (dismissed):

brayns/io/loaders/volume/VolumeLoader.cpp
NadirRoGue: Addressed. 
There is an unhandled corner case on the straem into which I ran (empty line at the end of the file). I think it would be nice to handle it since its pretty common.

brayns/io/loaders/volume/VolumeLoader.cpp
NadirRoGue: I think its superflous on this scenario, since the result is used to fill the Blob to load the volume using the RawLoader.

brayns/io/loaders/volume/VolumeLoader.cpp
NadirRoGue: `const char *` cannot be implicitly converted into `std::vector`

brayns/io/loaders/volume/VolumeLoader.cpp
NadirRoGue: I dont understand this one

python/brayns/core/loader/mhd_volume_loader.py
NadirRoGue: Addressed

python/brayns/core/loader/volume_data_type.py
NadirRoGue: Addressed

python/tests/core/loader/test_mhd_volume_loader.py
NadirRoGue: Addressed

Adrien4193 (approved):

PR #1144, opened 2023-01-13
BRAYNS 504 - Complete utils tests

NadirRoGue:


brayns/utils/image/ImageCodecRegistry.cpp
```diff
@@ -85,8 +88,10 @@ class ImageCodecStorage
 
 namespace brayns
 {
-const ImageCodec &ImageCodecRegistry::getCodec(const std::string &format)
+const ImageCodec &ImageCodecRegistry::getCodec(std::string format)
 {
+    StringCase::lower(format);
```
Adrien4193, line 92: I think this is misleading to make a copy of the string, we just want to read it, not take ownership of it.

You can use StringCase::toLower to make a lower copy of it.

brayns/utils/image/Image.cpp
```diff
@@ -81,7 +82,22 @@ size_t Image::getPixelSize() const
...
+    if (image.getChannelCount() != getChannelCount() || image.getChannelSize() != getChannelSize())
+    {
+        throw std::invalid_argument("Images have different shape");
+    }
+
+    if (image.getWidth() + x > getWidth() || image.getHeight() + y > getHeight())
+    {
+        throw std::out_of_range("Image write overflow");
+    }
+
```
Adrien4193, line 94: For me this is an invalid usage so I would just put an assert and not test that but it is ok like that also...

brayns/utils/image/ImageInfo.cpp
```diff
@@ -45,6 +47,11 @@ size_t ImageInfo::getPixelSize() const
 
 size_t ImageInfo::getPixelIndex(size_t x, size_t y) const
 {
+    if (x >= width || y >= height)
+    {
+        throw std::out_of_range("Image coordinates out of bounds");
+    }
```
Adrien4193, line 53: Or assert (like you prefer but I was thinking about it like for std::vector, we assume the user provides correct inputs to avoid runtime overhead in release mode).

brayns/utils/image/codecs/JpegCodec.cpp
```diff
@@ -32,6 +32,7 @@ std::string JpegCodec::getFormat() const
 
 std::string JpegCodec::encode(const Image &image, int quality) const
 {
+    quality = std::max(100 - std::max(quality, 0), 0);
```
Adrien4193, line 35: We already check that at the JSON-RPC level, so if the quality is invalid, it would be an error from our side no ?

In this case an assert would avoid hiding a bug.

tests/helpers/TemporaryFilename.h
```diff
@@ -0,0 +1,78 @@
...
+        return "/" + result + "/fake_file";
+    }
+
+    static std::string generateValid(size_t len = 10)
+    {
+        std::string result;
+        do
+        {
+            result = "/tmp/" + _generate(len);
+        } while (_exists(result));
```
Adrien4193, line 47: If you do directly std::filesystem::exists(result) it doesn't work ?

tests/unit/brayns/utils/TestDynamicLib.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+        CHECK_NOTHROW(brayns::DynamicLib(TestPaths::Libraries::braynsLibrary));
+    }
+    SUBCASE("Symbol address fetching")
+    {
+        auto lib = brayns::DynamicLib(TestPaths::Libraries::testLibrary);
+
+        CHECK(lib.getSymbolAddress("validExportedFunction"));
+        CHECK(!lib.getSymbolAddress("nonExportedFunction"));
+        CHECK(!lib.getSymbolAddress("nonExistingFunction"));
```
Adrien4193, line 42: CHECK_FALSE for people with bad eyes :)

Adrien4193 (dismissed):

brayns/utils/image/ImageCodecRegistry.cpp
NadirRoGue: I dont see the misleading part of that. Besides, this is not taking ownership of the string. Is making a copy out of it.

brayns/utils/image/Image.cpp
NadirRoGue: For me its an API, so it should be handled at runtime. The assert part is very specific to the usage we make of it in Brayns, but since it is not a performance critical part of the code (we do not even use it), I prefer to go with the first part which is a more correct design.

brayns/utils/image/ImageInfo.cpp
NadirRoGue: Same reply I made above.

brayns/utils/image/codecs/JpegCodec.cpp
NadirRoGue: The error here is that 0 is the lowest quality for JPG, while it is the maximum for PNG, so this fixes the parameter so it means the same for both formats.

tests/helpers/TemporaryFilename.h
NadirRoGue: I just put it on a separate static function to leave the main entry function of the class cleaner. After inlining, it will result in the same code you are suggesting.

tests/unit/brayns/utils/TestDynamicLib.cpp
NadirRoGue: Addressed

brayns/utils/image/ImageCodecRegistry.cpp
Adrien4193: That's what I say, with the copy in the signature, it looks like we are taking ownership of the string while we just read it (the lower copy is just an implementation detail, it shouldn't go in the function signature (imo).

But I did this before C++17, now the best would be string_view as we don't need specifically an std::string...

NadirRoGue: I still dont see the problem of having the copy in the signature, what is confusing about it?

brayns/utils/image/codecs/JpegCodec.cpp
Adrien4193: I don't think stbi uses quality for PNG but from the WebAPI we use 0 = worst and 100 = best no ?

So we should change that in the PngCodec, not the JPEG no ?

brayns/utils/image/ImageCodecRegistry.cpp
Adrien4193: As I said I would leave the copy as an implementation detail so we don't see it in the signature and it is clear the input is just something we are reading from (with const string & or string_view).

For me when I see an parameter expensive to copy passed by value, it means that I am transfering its ownership to the object I call (like with unique_ptr).

But that's just personal feeling, you can leave it like that if you prefer.

Adrien4193 (approved):

brayns/utils/image/codecs/JpegCodec.cpp
NadirRoGue: When Brayns was using FreeImage, that parameter has been used as PNG compression rate. 0 meant no compression, which gave bigger files and faster decoding.

NadirRoGue: The best solution IMO would be to rename the parameter to compression (as it is more accurate and clear). For some reason, however, stbi has the png compression ratio hardcoded as a global variable, but can be edited before every call to to the encoder.

In this specific case, the tests where failing. The reason is that JPG quality on stbi ranges from 1 to 100. When inputing 0 (default quality we use in Brayns), stbi uses by default 90, which reduces a bit the quality. This resulted on smaller files, but the ImageValidator failed to pass the test.

brayns/utils/image/ImageCodecRegistry.cpp
NadirRoGue: I believe the correct way is to leave like that, because the end goal is to make a copy of the string, thus the user is aware that will be done. Passing a const std::string & or an std::string_view hides this fact.

Furthermore, I believe your definition of "transfer ownership" is wrong. This would imply that the caller is loosing the object he is passing (thats why the definition works for an unique_ptr, its mean to be owned by in just one place, hence the need to cast it to an rvalue reference if its an lvalue one). In this case, it is a copy, and the caller still owns and can manipulate as he pleases his object.

PR #1145, opened 2023-01-13
Brayns 501 add network unit tests

Adrien4193:


brayns/network/websocket/WebSocket.cpp
```diff
@@ -80,36 +105,61 @@ class WebSocketSender
...
+size_t WebSocket::getId() const
 {
-    WebSocketBuilder::build(_socket);
-}
-
-WebSocket::WebSocket(Poco::Net::HTTPServerRequest &request, Poco::Net::HTTPServerResponse &response)
-    : _socket(request, response)
-{
-    WebSocketBuilder::build(_socket);
+    return reinterpret_cast<size_t>(this);
```
NadirRoGue, line 162: std::uintptr_t ?

tests/unit/brayns/network/TestEntrypoint.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include <doctest/doctest.h>
+
+#include <brayns/network/entrypoint/EntrypointBuilder.h>
+#include <brayns/network/entrypoint/EntrypointFinder.h>
+
+#include "MockEntrypoint.h"
+#include "MockInterface.h"
+#include "MockWebSocket.h"
+
+brayns::RequestMessage mockMessage()
```
NadirRoGue, line 29: Class with a static method instead of a free function?

brayns/network/websocket/WebSocket.cpp
Adrien4193: Well using the address as the ID is an implementation detail, from the caller point of view, it is supposed to be an unsigned integer and I used size_t so we can use it as an index.

In any case they both typedef to the same underlying type unless we have weird old segmented architecture.

tests/unit/brayns/network/TestEntrypoint.cpp
Adrien4193: Yeah I was the one pushing for that :)

NadirRoGue (dismissed):

NadirRoGue (approved):

PR #1153, opened 2023-01-27
BRAYNS 516 - Instantiation models in batch

NadirRoGue:


python/brayns/core/model.py
```diff
@@ -198,26 +198,34 @@ def get_models(instance: Instance) -> list[Model]:
...
+    if len(transforms) == 0:
+        raise ValueError("transforms list cannot be empty")
+
+    serialized_transforms = list[dict[str, Any]]()
+    for transform in transforms:
+        serialized_transform = serialize_transform(transform)
+        serialized_transforms.append(serialized_transform)
+
+    params: dict[str, Any] = {"model_id": model_id, "transforms": serialized_transforms}
+
```
Adrien4193, line 226:
```
params = {
    "model_id": model_id,
   "transforms": [serialize_transform(transform) for transform in transforms],
}
```

NadirRoGue: I think its fine either way. It can be changed later if needed.

python/brayns/core/model.py
```diff
@@ -198,26 +198,34 @@ def get_models(instance: Instance) -> list[Model]:
...
+    instances = instance.request("instantiate-model", params)
+    return [deserialize_model(instanceModel) for instanceModel in instances]
```
Adrien4193, line 228: Use snake_case instead of camelCase.

I would just name it model and models btw (just to avoid confusion with brayns.Instance).

```return [deserialize_model(model) for model in models]```

Adrien4193 (approved):

python/brayns/core/model.py
NadirRoGue: Addressed

Adrien4193 (approved):

PR #1217, opened 2023-11-24
BRAYNS-595 Refactor network loop.

Adrien4193:


tolokoban (approved):
LGTM as the young generation would say!

PR #1218, opened 2023-12-09
deps: bump spdlog to be in sync with future Spack

matz-e:


Adrien4193:
Hello,

We cannot merge to master, I will make a PR to develop to bump all dependencies so we are up-to-date.

Is this actually necessary for building with spack ? Couldn't we just update the receipe ?

matz-e:
> Hello,
> 
> We cannot merge to master, I will make a PR to develop to bump all dependencies so we are up-to-date.
> 
> Is this actually necessary for building with spack ? Couldn't we just update the receipe ?

I can do that beforehand (already did for next deployment). As these C++ packages move forward, it would be nice if the sources sort-of-followed at some point.

Adrien4193:
See https://github.com/BlueBrain/Brayns/pull/1219

PR #1220, opened 2023-12-13
BRAYNS-596 Fix morphology model.

Adrien4193:


tolokoban (approved):

jdcourcol:
@Adrien4193 @tolokoban I think this change needs some tests.

Adrien4193:
> @Adrien4193 @tolokoban I think this change needs some tests.

Yeah for now it is only on develop anyway.

For now I tested with morphology files that I hacked to reproduce the different soma types but if there are "official" samples, I can add them in the automated testing.

PR #1222, opened 2023-12-15
cleanup unused headers

mgeplf:


mgeplf:
I'm working on the error handling in morphio, and I noted that the above headers aren't used (at least, it seems to build in spack)

PR #1223, opened 2023-12-18
BRAYNS-599 Fix default cell orientations.

Adrien4193:


tolokoban (approved):
Bold move, but it does the trick.
Congrats!

PR #1288, opened 2024-11-22
Include missing cstdint header, bump spdlog

matz-e:
Noticed on Ubuntu 24 with GCC 13. spdlog needed to disambiguate some format issues.

Latest Spack builds clean with this PR applied.

### DIFF
--- a/brayns/utils/string/StringParser.h
+++ b/brayns/utils/string/StringParser.h
@@ -21,6 +21,7 @@
 
 #pragma once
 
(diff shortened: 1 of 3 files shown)


Adrien4193:
Hi Matthias, thanks for the PR.
We don't push directly to master, I included your changes to the develop branch and made a release.
Here is the spack PR:
https://github.com/BlueBrain/spack/pull/2527

